PR #2, opened 2018-03-30
Config updates

sbsmith86:
Some updates on how things were configured. 

* Lists out all of the scopes my client should have when configuring the northstar service.
* Defines an `importer` queue that all jobs from this app will be placed on
* Configures Laravel Horizon to run that queue
* Tags queue jobs so we know what is running when

image

### DIFF
--- a/app/Jobs/ImportTurboVotePosts.php
+++ b/app/Jobs/ImportTurboVotePosts.php
@@ -33,6 +33,17 @@ class ImportTurboVotePosts implements ShouldQueue
         $this->filepath = $filepath;
     }
 
+
+    /**
+     * Get the tags that should be assigned to the job.
+     *
+     * @return array
+     */
+    public function tags()
+    {
+        return ['turbovote'];
+    }
+
     /**
      * Execute the job.
      *
--- a/config/services.php
+++ b/config/services.php
@@ -65,7 +65,7 @@ return [
         'client_credentials' => [
             'client_id' => env('NORTHSTAR_CLIENT_ID'),
             'client_secret' => env('NORTHSTAR_CLIENT_SECRET'),
-            'scope' => ['user', 'admin'],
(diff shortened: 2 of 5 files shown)


app/Jobs/ImportTurboVotePosts.php
```diff
@@ -33,6 +33,17 @@ public function __construct($filepath)
...
 
+
+    /**
+     * Get the tags that should be assigned to the job.
+     *
+     * @return array
+     */
+    public function tags()
+    {
+        return ['turbovote'];
```
sbsmith86: Thinking we should add some sort of date to the tag for better 👀 but this is just a default for now.

katiecrane: @sbsmith86 what is this used for?

sbsmith86: @katiecrane it allows us to "tag" queued jobs so if you wanted some unique identifier for a job put on a queue you would configure that here. 

Like we can tag these jobs as all `turbovote` jobs, a general signup import job could be `signup` or whatever. And then you can monitor the tag in Horizon. Still working out the usefulness of this, but I wanted to see it in action.

katiecrane: @sbsmith86 oh cool! that definitely sounds useful!

config/services.php
```diff
@@ -65,7 +65,7 @@
         'client_credentials' => [
             'client_id' => env('NORTHSTAR_CLIENT_ID'),
             'client_secret' => env('NORTHSTAR_CLIENT_SECRET'),
-            'scope' => ['user', 'admin'],
+            'scope' => ['admin', 'user', 'client', 'activity', 'write'],
```
DFurnes: @sbsmith86 Before we deploy this to production, let's remove the `client` scope! That allows an application to read/write OAuth client details, which we don't want most applications (aside from Aurora) to be able to do.

PR #14, opened 2018-04-27
Create Northstar Users On the Go

sbsmith86:
#### What's this PR do?

Overall this refactors the `ImportTurboVotePosts` job so the code is easier to follow (I hope). If it is hard to follow, I suggest viewing the PR in "Split" mode to remove any refactor noise. 

Mainly - This includes an update to the logic that will create a northstar user if we do not have a Northstar Id passed to us in the referral code. 

I am testing my results against the results that Sohaib expects to get using his analysis. So far we match up except for missing NS users.

Also I would love for more eyes on the logic in the `getOrCreateUser()` method because I think this could be cleaner, but also does it make sense?

#### Relevant Tickets

Chompy things.

### DIFF
--- a/app/Jobs/ImportTurboVotePosts.php
+++ b/app/Jobs/ImportTurboVotePosts.php
@@ -91,9 +91,9 @@ class ImportTurboVotePosts implements ShouldQueue
                     if ($user) {
                         // @TODO - We probably don't need to do this for new users and can skip this.
                         $post = $rogue->getPost([
-                                'campaign_id' => (int) $referralCodeValues['campaign_id'],
-                                'northstar_id' => $user->id,
-                                'type' => 'voter-reg',
+                            'campaign_id' => (int) $referralCodeValues['campaign_id'],
+                            'northstar_id' => $user->id,
+                            'type' => 'voter-reg',
                         ]);
 
                         if (! $post['data']) {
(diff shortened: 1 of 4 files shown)


app/Jobs/ImportTurboVotePosts.php
```diff
@@ -340,5 +320,63 @@ private function scrubRecord($record)
...
+            $user = gateway('northstar')->asClient()->createUser([
+                'email' => $record['email'],
+                'mobile' => $record['phone'],
+                'first_name' => $record['first-name'],
+                'last_name' => $record['last-name'],
+                'addr_street1' => $record['registered-address-street'],
+                'addr_street2' => $record['registered-address-street-2'],
+                'addr_city' => $record['registered-address-city'],
+                'addr_state' => $record['registered-address-state'],
+                'addr_zip' => $record['registered-address-zip'],
```
katiecrane: Do you need to check if these fields are set like you do in all the conditionals above?

app/Jobs/ImportTurboVotePosts.php
```diff
@@ -340,5 +320,63 @@ private function scrubRecord($record)
...
+    }
+
+    private function getCSVRecords($filepath)
+    {
+        $file = Storage::get($filepath);
+        $csv = Reader::createFromString($file);
+        $csv->setHeaderOffset(0);
+        $records = $csv->getRecords();
+
+        event(new LogProgress('Total rows to chomp: ' . count($csv)));
```
katiecrane: I love this

app/Jobs/ImportTurboVotePosts.php
```diff
@@ -340,5 +320,63 @@ private function scrubRecord($record)
...
+    /**
+     * For a given record and referral code values, first check if we have a northstar ID, then grab the user using that.
+     * Otherwise, see if we can find the user with the given email (if it exists), if not check if we can find them with the given phone number (if it exists).
+     * If all else fails, create the user .
+     *
+     * @TODO - If we have a northstar id in the referral code, then we probably don't need to make a call to northstar for the full user object.
+     *
+     * @return array
+     */
+    private function getOrCreateUser($record, $values)
```
katiecrane: I think that this method makes sense. Nothing really jumps out at me as being unclear either!

app/Jobs/ImportTurboVotePosts.php
sbsmith86: these aren't required. so i didn't do the extra check.

sheyd:
For a second I read this PR title as Create Northstar Users in Go, and thought we were migrating Chompy to Go-pher

😂 and got all excited.

PR #15, opened 2018-05-09
If the referral code is empty, don't skip the record

sbsmith86:
#### What's this PR do?
This updates the chompy logic that imports records from a TurboVote export. 

Before, we would loop through each record and skip any records in the CSV that did not have a `referral-code`. 

After some tests, we decided that we actually do _not_ want to skip these records. 

If the referral code is empty, we continue to process the record, use a standard set of default values, and then create or get the user from northstar using the `email` or `phone` provided. 

The core of the work is [here ](https://github.com/DoSomething/chompy/compare/count-empty-referralcodes?diff=split&expand=1&name=count-empty-referralcodes#diff-ed45aaf47afdecacbc92e2d0393dc00bL183) and that is just the method that decides how to parse the referral code or if it doesn't exist, set the default referral code values. 

👀 

This might be hard to view because I had indetation that needed to change. I would look at this in split view for easier reading. 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/157411893) Pivotal Card.

#### How to deploy
https://github.com/DoSomething/rogue/blob/master/docs/development/deployments.md

### DIFF
--- a/app/Jobs/ImportTurboVotePosts.php
+++ b/app/Jobs/ImportTurboVotePosts.php
@@ -216,7 +216,6 @@ class ImportTurboVotePosts implements ShouldQueue
             ];
         }
 
-
         return $values;
     }
 


app/Jobs/ImportTurboVotePosts.php
```diff
@@ -345,7 +349,7 @@ private function getOrCreateUser($record, $values)
         foreach ($userFieldsToLookFor as $field => $value)
         {
             if ($value) {
-                info('getting user with the '.$field.' field', [$field => $value]);
+                // info('getting user with the '.$field.' field', [$field => $value]);
```
katiecrane: Should this be either uncommented or removed?

sbsmith86: ah I just kept it there because It was nice for development without having to add to many flags.

app/Jobs/ImportTurboVotePosts.php
```diff
@@ -184,35 +174,49 @@ private function parseReferralCode($referralCode)
...
+            $values = [
+                'northstar_id' => null, // set the user to null so we force account creation when the code is not present.
+                'campaign_id' => 8017,
+                'campaign_run_id' => 8022,
+                'source' => 'turbovote',
+                'source_details' => null,
+            ];
         }
 
+
```
katiecrane: I think this is an extra blank line?

sbsmith86: sure is! 👾

app/Jobs/ImportTurboVotePosts.php
katiecrane: Oooo is this log not something that would be helpful on production?

sbsmith86: not really, gateway will log all the requests made anyway so we got logs like 

```
[2018-05-09 18:37:06] local.INFO: progress_log: Processing: 5a84b518-2466-48ab-b013-8fb90c90c1c9  
[2018-05-09 18:37:06] local.INFO: Request made. {"method":"GET","uri":"http://northstar.test/v1/users/email/thatonegirllaila@gmail.com","request_id":"b2e06fb0-9a62-4c72-af11-62f6d485ed23"} 
[2018-05-09 18:37:06] local.INFO: Request made. {"method":"GET","uri":"http://northstar.test/v1/users/mobile/16087180905","request_id":"f241a99b-cde4-46be-ab84-83dbae5ddd65"} 
[2018-05-09 18:37:07] local.INFO: Request made. {"method":"POST","uri":"http://northstar.test/v1/users","request_id":"5fdfc533-cf43-4ef6-bc7c-bc170dea9dde"} 
[2018-05-09 18:37:07] local.INFO: created user {"user":"5af33fd39a892006bf539c92"} ```

and it is sort of easy to see the different ways we are trying to grab the user from NS

katiecrane: cool! leaving this commented out is fine with me then!

sbsmith86:
@katiecrane you are a 🌟 THANK YOU

PR #16, opened 2018-05-09
Validate referral code

sbsmith86:
### What does this PR do

We use the `referral-code` column in the turbovote csv to associate user/campaign data with the voter-reg post. 

Mostly, the `referral-code` looks like: `user:5a854f70a0bfad749a0449f1,campaignID:8017,campaignRunID:8022,source:email,source_details:newsletter_266` 
This is good. 

But sometimes the `referral-code` looks like: 
`11_facts,utm_source:dosomething,utm_medium:11_facts,utm_campaign:8017,utm_content:bullying`

Not as good. 

So this updates the logic to make sure we have what we need, otherwise use the defaults.

### DIFF
--- a/app/Jobs/ImportTurboVotePosts.php
+++ b/app/Jobs/ImportTurboVotePosts.php
@@ -207,7 +207,7 @@ class ImportTurboVotePosts implements ShouldQueue
             }
         }
 
-        // If the referral code is not present or we didn't get the right fields, use these default values.
+        // Make sure we have all the values we need, otherwise, use the defaults.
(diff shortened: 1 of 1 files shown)


katiecrane:
@sbsmith86 in the example you gave, would `utm_campaign:8017` mean you'd want the post to have `campaign_id` of `8017`?

sbsmith86:
@katiecrane i guess, but since the default is `8017` doesn't seem like we need to account for a `utm` parameter.

PR #18, opened 2018-06-04
Kill horizon

sbsmith86:
#### What's this PR do?

This PR removes the laravel horizon package in favor of just running the queue directly. This package was really cool, and I could see a lot of use for it. But it doesn't seem to handle long running queue jobs well. 

Mainly, jobs would die after a certain timeout that was set in the horizon configuration. So I would set that timeout to be higher, but then the horizon interface would not update until that timeout was over. 

So, there could be an instance where I set the timeout to be 30 mins to account for long running jobs. But, a smaller job would finish in like 20 mins and the horizon interface would not update to indicate the status of the job was done for 30mins. This caused confusion in checking the status of the job. 

I can see a use for horizon for smaller queue workers, but it doesn't seem great for what chompy is doing since we need to process large CSVs.

👀


DFurnes:
Aw man, that's a bummer! 😢

PR #23, opened 2018-06-21
Setup command environmentalist

deguzmankevin:
This PR adds documentation and helps you set up Chompy the exact same way you set up our other apps! Shoutout to @DFurnes for environmentalist!!


app/Console/Commands/SetupCommand.php
```diff
@@ -0,0 +1,101 @@
...
+            $environments = [
+                'https://rogue-qa.dosomething.org' => 'https://aurora-qa.dosomething.org',
+                'https://rogue-thor.dosomething.org' => 'https://aurora-thor.dosomething.org',
+            ];
+
+            $this->chooseEnvironmentVariable('Rogue_URL', 'Choose a Rogue environment', array_keys($environments));
+
+        });
+
+        $this->section('Set Driver environment variables', function () {
```
sbsmith86: We don't need this section, in general the defaults are fine.

app/Console/Commands/SetupCommand.php
```diff
@@ -0,0 +1,101 @@
...
+        });
+
+        $this->section('Set Pusher environment variables', function () {
+            $this->setEnvironmentVariable('PUSHER_APP_ID', 'Enter the Pusher App ID');
+            $this->setEnvironmentVariable('PUSHER_APP_KEY', 'Enter the Pusher App Key');
+            $this->setEnvironmentVariable('PUSHER_APP_SECRET', 'Enter the Pusher App Secret');
+
+        });
+
+        $this->section('Set S3 environment variables', function () {
```
sbsmith86: I think we can remove this, and maybe add a comment that the `FILESYSTEM_DRIVER` will determine if S3 is used to store files or if the local storage driver would be used.

deguzmankevin:
thanks for all the suggestions! fixed! @DFurnes @sbsmith86 :raised_hands:

PR #30, opened 2018-07-03
Update setup command with new Aurora URLs.

DFurnes:
This pull request updates Aurora URLs in the setup command, clarifies the difference between auth code & client credentials environment variables a bit, and removes `FILESYSTEM_DRIVER` from the wizard since we should always be using `local` here.

References DoSomething/devops#388.


DFurnes:
I've pre-selected `https://identity-qa.dosomething.org` as the default value for Northstar (so you can just hit enter, enter, enter as you go through the setup process. Lemme know if that's not what people are using and I can switch it to one of the others!

PR #31, opened 2018-07-09
Add TrustProxies & ForceHttps middleware.

DFurnes:
This pull request adds middleware to trust proxies (so we know if we're on HTTPS or not), and then to force HTTPS and the correct canonical URL.

### DIFF
--- a/app/Http/Middleware/ForceHttps.php
+++ b/app/Http/Middleware/ForceHttps.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Chompy\Http\Middleware;
+
+use Closure;
+use League\Uri\Schemes\Http;
+use function League\Uri\parse;
+
+class ForceHttps
+{
+    /**
+     * Handle an incoming request.
+     *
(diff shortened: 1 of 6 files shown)


app/Http/Middleware/ForceHttps.php
```diff
@@ -0,0 +1,39 @@
...
+     */
+    public function handle($request, Closure $next)
+    {
+        // If running in a production environment, redirect all traffic to HTTPS
+        // and the correct canonical URL, e.g. never '*.herokuapp.com'!
+        if (config('app.env') == 'production') {
+            $canonicalHost = parse(config('app.url'))['host'];
+            $hasIncorrectHost = $request->header('Host') !== $canonicalHost;
+
+            if ($hasIncorrectHost || ! $request->secure()) {
```
katiecrane: Aw I see now that you made this change from `$isInsecure` to `! $request->secure()` in Aurora too! I dig it!

PR #32, opened 2018-07-31
Import Rock The Vote posts/signups to Rogue

katiecrane:
I've finally tested this enough locally to merge and do some testing on QA.

This includes:
1. The job to import Rock The Vote information into Rogue. This is based on [this detailed write-up](https://github.com/DoSomething/chompy/wiki/How-Rock-The-Vote-(RTV)-records-are-imported) and I can provide an example file if you'd like one.

2. Adds a checkbox for the Rock The Vote import.

3. Removes `client` from the scopes we get from Northstar, we don't need it!

This is a big pile of code and we are going to test thoroughly on QA, but please let me know if you have any questions or if there is anything I can explain better! It gets pretty gritty but I tried to include some helpful comments.

### DIFF
--- a/app/Jobs/ImportRockTheVotePosts.php
+++ b/app/Jobs/ImportRockTheVotePosts.php
@@ -185,29 +185,34 @@ class ImportRockTheVotePosts implements ShouldQueue
                     $value = explode('=', $value);
                 }
 
-                // Grab northstar id
-                if (strtolower($value[0]) === 'user') {
-                    $values['northstar_id'] = $value[1];
-                }
+                // Add northstar_id, campaign id and run, source, and source details into $values
+                switch (strtolower($value[0])) {
+                    case 'user':
+                        $values['northstar_id'] = $value[1];
+                        break;
 
-                // Grab the Campaign Id.
-                if (strtolower($value[0]) === 'campaignid' || strtolower($value[0]) === 'campaign') {
-                    $values['campaign_id'] = $value[1];
-                }
+                    case 'campaignid':
+                        $values['campaign_id'] = $value[1];
+                        break;
 
-                // Grab the Campaign Run Id.
-                if (strtolower($value[0]) === 'campaignrunid') {
-                    $values['campaign_run_id'] = $value[1];
-                }
+                    case 'campaign':
+                        $values['campaign_id'] = $value[1];
+                        break;
 
-                // Grab the source
-                if (strtolower($value[0]) === 'source') {
-                    $values['source'] = $value[1];
-                }
+                    case 'campaignrunid':
+                        $values['campaign_run_id'] = $value[1];
+                        break;
 
-                // Grab any source details
-                if (strtolower($value[0]) === 'source_details') {
-                    $values['source_details'] = $value[1];
+                    case 'source':
+                        $values['source'] = $value[1];
+                        break;
+
+                    case 'source_details':
+                        $values['source_details'] = $value[1];
+                        break;
+
+                    default:
+                        break;
                 }
 
                 // Is this a referral?
--- a/app/Jobs/ImportRockTheVotePosts.php
+++ b/app/Jobs/ImportRockTheVotePosts.php
@@ -62,7 +62,7 @@ class ImportRockTheVotePosts implements ShouldQueue
      */
     public function handle(Rogue $rogue)
     {
-        info('STARTING');
+        info('STARTING RTV IMPORT');
 
         $records = $this->getCSVRecords($this->filepath);
 
@@ -71,7 +71,7 @@ class ImportRockTheVotePosts implements ShouldQueue
             $shouldProcess = $this->scrubRecord($record);
 
             if ($shouldProcess) {
-                // info('progress_log: Processing: ' . $record['id']); @TODO: do we need this? no nsid column
+                info('progress_log: Processing: ' . $record['Email address']);
 
                 $referralCode = $record['Tracking Source'];
                 $referralCodeValues = $this->parseReferralCode($referralCode);
@@ -79,7 +79,7 @@ class ImportRockTheVotePosts implements ShouldQueue
                 try {
                     $user = $this->getOrCreateUser($record, $referralCodeValues);
                 } catch (\Exception $e) {
-                    info('There was an error with that user: ' . $record['id'], [
+                    info('There was an error with that user: ' . $record['Email address'], [
                         'Error' => $e->getMessage(),
                     ]);
                 }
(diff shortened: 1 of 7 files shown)


resources/views/pages/import.blade.php
```diff
@@ -21,8 +21,10 @@
             <div class="form-check">
                 <label class="form-check-label">
                     <!-- @TODO - make "checked" status a variable that has a default -->
-                    <input checked name="import-type" class="form-check-input" type="checkbox" value="turbovote">
+                    <input name="import-type" class="form-check-input" type="checkbox" value="turbovote">
                     TurboVote Import
```
chloealee: this is totally nitpicky but maybe we can add a `<br>` here so each checkbox is on its own line?

app/Jobs/ImportRockTheVotePosts.php
```diff
@@ -0,0 +1,459 @@
...
+            $user = gateway('northstar')->asClient()->createUser($userData);
+
+            info('created user', ['user' => $user->id]);
+            $this->stats['countUserAccountsCreated']++;
+        }
+
+        return $user;
+    }
+
+    private function getCSVRecords($filepath)
```
chloealee: question! I noticed that this is also private in the [ImportTurboVotePosts Job](https://github.com/DoSomething/chompy/blob/master/app/Jobs/ImportTurboVotePosts.php#L387) as well and think I will need to use this exact same function again when I write the Facebook Share Import Job. Is there a reason this has to be private or can we make this a helper?

app/Jobs/ImportRockTheVotePosts.php
```diff
@@ -0,0 +1,459 @@
...
+
+                // Grab any source details
+                if (strtolower($value[0]) === 'source_details') {
+                    $values['source_details'] = $value[1];
+                }
+
+                // Is this a referral?
+                if (strtolower($value[0]) === 'referral' && strtolower($value[1]) === 'true') {
+                    $referral = true;
+                }
```
chloealee: I think this is totally preference but this could be a nice place to use a [switch](http://php.net/manual/en/control-structures.switch.php) statement?

app/Jobs/ImportRockTheVotePosts.php
```diff
@@ -0,0 +1,459 @@
...
+    {
+        return [
+            'totalRecords' => 0,
+            'countScrubbed' => 0,
+            'countProcessed' => 0,
+            'countPostCreated' => 0,
+            'countUserAccountsCreated' => 0,
+        ];
+    }
+
```
chloealee: same with this! Is this possible to pull out to a helper?

app/Jobs/ImportRockTheVotePosts.php
```diff
@@ -0,0 +1,459 @@
...
+     * @return bool
+    */
+    private function scrubRecord($record)
+    {
+        $isNotValidEmail = strrpos($record['Email address'], 'thing.org') !== false || strrpos($record['Email address'] !== false, '@dosome') || strrpos($record['Email address'], 'rockthevote.com') !== false || strrpos($record['Email address'], 'test') !== false || strrpos($record['Email address'], '+') !== false;
+        $isNotValidLastName = strrpos($record['Last name'], 'Baloney') !== false;
+
+        return $isNotValidEmail || $isNotValidLastName ? false : true;
+    }
+
```
chloealee: and this (can we make a helper)?

app/Jobs/ImportRockTheVotePosts.php
katiecrane: @chloealee great question! I was looking into this but a lot of the functions had to be further customized, but I don't think that this one does! I'll make sure they match and pull it out!

app/Jobs/ImportRockTheVotePosts.php
katiecrane: good call!!

app/Jobs/ImportRockTheVotePosts.php
katiecrane: @chloealee this one differs between Rock the Vote and Turbovote so I think it's gotta stay separate

app/Jobs/ImportRockTheVotePosts.php
katiecrane: Okay so @chloealee since we're modifying instance variables in this one I decided to add this and `initStats()` as a trait called `ImportToRogue`. I don't _love_ traits because I think it's confusing to follow when you call `$this->{whatever}` and then the function is not in that file, but in this case hopefully it will help DRY things up and make Chompy overall easier to work with. We can always revert back! I also did not update the turbovote file because I'm not able to test that currently and we don't have any tests written .

app/Jobs/ImportRockTheVotePosts.php
katiecrane: @chloealee yes I like this way better!!!!

resources/views/pages/import.blade.php
katiecrane: @chloealee added!

app/Jobs/ImportRockTheVotePosts.php
chloealee: @katiecrane totally agree with you that sometimes traits can get a little confusing but love this solution here that you did! If it gets too hairy, we can change back like you said! This is super helpful though bc I definitely use those functions as well and we can now cut a lot of code in the jobs :)

app/Traits/ImportToRogue.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace Chompy\Traits;
+
+use League\Csv\Reader;
+use Chompy\Events\LogProgress;
+use Illuminate\Support\Facades\Storage;
+
+trait ImportToRogue
```
chloealee: @katiecrane do we have to add this in the `use` statement [here](https://github.com/DoSomething/chompy/pull/32/files#diff-ed45aaf47afdecacbc92e2d0393dc00bR19)?

chloealee: sorry disregard! The changes were hiding there and saw you already did that ! :)

app/Jobs/ImportTurboVotePosts.php
```diff
@@ -43,7 +43,7 @@ public function __construct($filepath)
     {
         $this->filepath = $filepath;
 
-        $this->stats = $this->statsInit();
+        $this->stats = statsInit();
```
DFurnes: Sorry I'm late to the game, but why was `$this->` removed here. Is this calling the method on the trait?

app/Traits/ImportToRogue.php
DFurnes: I'm having trouble finding this too! Where was the trait being used?

app/Jobs/ImportTurboVotePosts.php
katiecrane: @DFurnes ah shoot! this is outdated from when I was trying it as a helper, I'll push up a fix now!

PR #33, opened 2018-08-01
Re-add removed code

katiecrane:
I was messing around with some things in `ImportTurbovotePosts` and accidentally included the changes in my last PR! This just reverts it back to how it was before.

### DIFF
--- a/app/Jobs/ImportTurboVotePosts.php
+++ b/app/Jobs/ImportTurboVotePosts.php
@@ -43,7 +43,7 @@ class ImportTurboVotePosts implements ShouldQueue
     {
         $this->filepath = $filepath;
 
-        $this->stats = statsInit();
+        $this->stats = $this->statsInit();
     }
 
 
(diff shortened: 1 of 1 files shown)


app/Jobs/ImportTurboVotePosts.php
```diff
@@ -153,6 +153,22 @@ public function handle(Rogue $rogue)
...
+        return [
+            'totalRecords' => 0,
+            'countScrubbed' => 0,
+            'countProcessed' => 0,
+            'countPostCreated' => 0,
+            'countUserAccountsCreated' => 0,
+        ];
+    }
+
     /**
```
chloealee: I think I am confused! Why wouldn't this work to keep in the `ImportToRogue` trait?

katiecrane: @chloealee I left it in there too. I am just hesitant to make any changes to this file because there are no tests so we'd have to test manually and I'm not sure what I'm looking for in the results. Does that seem reasonable or nah?

chloealee: oh right right ! That totally makes sense! Sorry - was thinking this was for Rock the Vote!

katiecrane: ah no worries! I know a lot of work and testing and analysis and planning went into the TurboVote import so I have an "if it ain't broke don't fix it" view on it currently

chloealee: agreed!

katiecrane: @DFurnes done and done!

PR #34, opened 2018-08-01
Job to Import Historical FB Share Posts. 

chloealee:
#### What's this PR do?
- Creates job to import historical FB share posts based on the logic described [here](https://github.com/DoSomething/chompy/wiki/How-Historic-FB-Share-Records-are-Imported). 
  - When uploading with example CSV, here is an example of the posts created in Rogue:
screen shot 2018-08-01 at 2 58 24 pm
screen shot 2018-08-01 at 2 58 40 pm
@ngjo does this look right to you? 

- Updates "Type of Import" buttons to be radio buttons instead of checkboxes: 
screen shot 2018-08-01 at 2 59 23 pm

We need to be careful when running this to make sure we don't create dupe posts! After the job has run, be sure to check the logs for anything that says "There was an error storing the post for"... we'll want to _only_ fix and re-upload those - if we include other data, Rogue will save duplicate posts. More info. in the convo [here](https://dosomething.slack.com/archives/CA0N5FXND/p1533147475000047).

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156165836) Pivotal Card.

### DIFF
--- a/app/Jobs/ImportFacebookSharePosts.php
+++ b/app/Jobs/ImportFacebookSharePosts.php
@@ -6,6 +6,7 @@ use League\Csv\Reader;
 use Chompy\Services\Rogue;
 use Illuminate\Bus\Queueable;
 use Chompy\Events\LogProgress;
+use Chompy\Traits\ImportToRogue;
 use Illuminate\Queue\SerializesModels;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\Queue\InteractsWithQueue;
@@ -14,7 +15,7 @@ use Illuminate\Foundation\Bus\Dispatchable;
 
 class ImportFacebookSharePosts implements ShouldQueue
 {
-    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels, ImportToRogue;
 
     /**
      * The path to the stored csv.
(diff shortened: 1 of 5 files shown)


chloealee:
@ngjo two questions to double check!
1. Will all records for FB Share Imports have a NS id? Are there any that we'll have to scrub (i.e. any with @dosomething.org or with the last name "baloney"?)? 
2. Will these actions all be "quiz-share"?

ngjo:
`Will all records for FB Share Imports have a NS id? Are there any that we'll have to scrub (i.e. any with @dosomething.org or with the last name "baloney"?)?`

The file might have some FB shares w/o NS id, but if it's easier for me to just delete those out of the file before importing, I can do that! And there shouldn't be any email addresses in the FB file, just NS IDs as the identifiers, so no scrubbing needed!


`Will these actions all be "quiz-share"?`
Is this for the `action` column? If so, we're still working out exactly what would be in the `action` column, but whatever we land on, it won't necessarily be the same for all of them.

ngjo:
Doooyyy @chloealee I just looked at the screenshots! So yea, `action` won't always be `quiz-share` and we're still trying to determine what we need to fill that in w/ for the backfill. [Long Slack thread here](https://dosomething.slack.com/archives/C03T8SDJG/p1533064322000191) for more context.

I think we'll need to also include the `data.url` (what was shared) in the post details column so that @sohaibhasan can have that for reporting!

chloealee:
@ngjo thanks for the quick response! 

> The file might have some FB shares w/o NS id, but if it's easier for me to just delete those out of the file before importing, I can do that! 

I can update the script to just skip this record if there is no NS id! I can make sure it logs that we didn't import it because it didn't have a NS id too if that is helpful! 

> And there shouldn't be any email addresses in the FB file, just NS IDs as the identifiers, so no scrubbing needed!

oh duh! 🤦‍♀️ 

>  If so, we're still working out exactly what would be in the action column, but whatever we land on, it won't necessarily be the same for all of them.

Another 🤦‍♀️ ! We already discussed this - my bad! I'll update this to pull from the `action` column instead of what it is now (hard coded). 

> I think we'll need to also include the data.url (what was shared) in the post details column so that @sohaibhasan can have that for reporting!

Oops I missed that ! I'll update that [here](https://github.com/DoSomething/chompy/wiki/How-Historic-FB-Share-Records-are-Imported) as well! So it should be something like this in the post `details` column? 

`"details": {"platform": "facebook", "url": "data.url"},`

app/Jobs/ImportFacebookSharePosts.php
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Create a new job instance.
+     *
+     * @return void
+     */
+    public function __construct($filepath)
+    {
+        $this->filepath = $filepath;
+
+        $this->stats = $this->statsInit();
```
katiecrane: @chloealee Hmmm now I'm wondering if this shouldn't be part of a trait and should be per import, because 2 of the 5 stats do not apply to this import, which doesn't hurt anything but they'll just sit at 0. What do you think?

chloealee: @katiecrane that is a good call! I don't have strong opinions but I don't think it hurts for them to sit at 0, mostly because that is still an accurate count (0 records needed to be scrubbed, 0 accounts needed to be made)?

katiecrane: @chloealee perfect! that's super reasonable to me!

PR #35, opened 2018-08-02
Add status to post details for RTV

katiecrane:
We just want to add another field ("Status") to the `details` on the post when creating a new post in Rogue. We decided that it makes more sense to include this as well as "Finish with State" because those are the two fields used to determine the overall `status` of the post.


katiecrane:
cc: @sheyd

PR #38, opened 2018-08-08
updates how we send created_at to Rogue

chloealee:
#### What's this PR do?
Sends `created_at` time in correct format that we want to send to Rogue. 

👀 

#### Relevant tickets
[This](https://www.pivotaltracker.com/n/projects/2019429/stories/156165836) Pivotal Card and this PR in Rogue.

### DIFF
--- a/app/Jobs/ImportFacebookSharePosts.php
+++ b/app/Jobs/ImportFacebookSharePosts.php
@@ -69,7 +69,7 @@ class ImportFacebookSharePosts implements ShouldQueue
         foreach ($records as $offset => $record) {
             info('process_log: Processing: ' . $record['_id']);
             $postData = [
-                'campaign_id' => (int) $record['campaign_id'],
+                'campaign_id' => $record['campaign_id'],
                 'campaign_run_id' => (int) $record['campaign_run'],
                 'northstar_id' => $record['user.northstarId'],
                 'type' => 'share-social',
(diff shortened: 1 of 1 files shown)


app/Jobs/ImportFacebookSharePosts.php
```diff
@@ -78,9 +78,8 @@ public function handle(Rogue $rogue)
                 'status' => 'accepted',
                 'source' => 'importer-client',
                 'source_details' => json_encode(['original-source' => $record['event.source']]),
-                'created_at' => strtotime($record['to_timestamp']),
+                'created_at' => date(DATE_ATOM, strtotime($record['to_timestamp'])),
```
katiecrane: @chloealee what's the `to_timestamp` column look like?

chloealee: @katiecrane here's an example: `6/8/2018 18:27:00`

chloealee: ohhh @katiecrane maybe I should do something like you and @sbsmith86 did in your imports:
```
strtolower(Carbon::parse($record['Started registration'])->format('F-Y'));
```

chloealee: oh wait, I think I am wrong I don't think we want to format that like that here.

katiecrane: I think you might be able to do something like this instead if you want to use Carbon:
```
Carbon::parse($record['to_timestamp'])->toDateTimeString();
```
Does that give you the right format?

chloealee: ah yes it does! Thanks @katiecrane ! Updated - I like that way better :)

PR #39, opened 2018-08-09
adds `_id` to details json in Historic FB Shares

chloealee:
#### What's this PR do?
adds `_id` to `details` json in Historic FB Shares so it will look like this:
```
{
  'platform': 'facebook',
  '_id': 5b201dfdeb77630004790a85,
}
```

👀 

From request in convo [here](https://dosomething.slack.com/archives/CA0N5FXND/p1533826403000235?thread_ts=1533755010.000294&cid=CA0N5FXND)

#### Relevant tickets
Continued work from #34

### DIFF
--- a/app/Jobs/ImportFacebookSharePosts.php
+++ b/app/Jobs/ImportFacebookSharePosts.php
@@ -75,7 +75,7 @@ class ImportFacebookSharePosts implements ShouldQueue
                 'northstar_id' => $record['user.northstarId'],
                 'type' => 'share-social',
                 'action' => $record['action'],
-                'details' => json_encode(['platform' => 'facebook']),
+                'details' => json_encode(['platform' => 'facebook', '_id' => $record['_id']]),
                 'status' => 'accepted',
                 'source' => 'importer-client',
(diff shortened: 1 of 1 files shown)


DFurnes:
What's `_id` in this context? User ID? Not sure if it's too late, but should we name it that instead?

chloealee:
@DFurnes this is the puck id! This is what it is named in the spreadsheet that we import to Chompy so we just named it the same thing here. @sohaibhasan thoughts?

sohaibhasan:
@chloealee @DFurnes , Puck sends us ```

chloealee:
@sohaibhasan @DFurnes should I update this to be:
`'details' => json_encode(['platform' => 'facebook', 'puck_id' => $record['_id']]),`

sohaibhasan:
That would be easier on the eyes, but not a big deal either way

chloealee:
Happy to do that to avoid future confusion! I'll set up another PR for that - thanks for flagging @DFurnes !

DFurnes:
Ah, that makes sense now. Thanks @chloealee!

PR #42, opened 2018-08-20
WIP - ability to upload CSV's into Chompy that come from Excel 

chloealee:
#### What's this PR do?

tell us, what have you changed?

tell us, how can we review, to test that this works

what else?

#### Relevant tickets

Fixes #

### DIFF
--- a/app/Jobs/ImportTurboVotePosts.php
+++ b/app/Jobs/ImportTurboVotePosts.php
@@ -390,6 +390,14 @@ class ImportTurboVotePosts implements ShouldQueue
         $file = Storage::get($filepath);
         $csv = Reader::createFromString($file);
         $csv->setHeaderOffset(0);
+        $csv = str_replace("\r","\n", $csv);
+        // dd($csv);
+
+        // // $cvs = array_map('trim', $csv->getHeader());
+        // // dd($headers);
+        // foreach ($csv->getRecords($headers) as $record) {
+        //     dd('hi');
+        // }
         $records = $csv->getRecords();
         $this->totalRecords = count($csv);
 


chloealee:
@DoSomething/team-bleed having a bit of trouble with this! I read that this is a known issue with [weird formatting](Chompy isn't able to upload CSVs that come from Excel because of this [known issue](https://nicercode.github.io/blog/2013-04-30-excel-and-line-endings/) from Excel. I tried this solution but then read some more about this on stackoverflow and seems like it is sort of a buggy solution. 

I am trying to go with @DFurnes 's suggestion in the [Pivotal Card](https://www.pivotaltracker.com/story/show/155369328/comments/192874327) and think I got it to look like [this](http://cl.ly/2w3c1Z1S2Q0r) however, it is a string and not an array as it is supposed to be:

screen shot 2018-08-20 at 4 56 14 pm when it gets passed to [`getRecords`](https://github.com/thephpleague/csv/blob/master/src/Reader.php#L245) [here]().

app/Jobs/ImportTurboVotePosts.php
```diff
@@ -390,7 +390,15 @@ private function getCSVRecords($filepath)
         $file = Storage::get($filepath);
         $csv = Reader::createFromString($file);
         $csv->setHeaderOffset(0);
-        $records = $csv->getRecords();
+        // $csv = str_replace("\r","\n", $csv);
```
DFurnes: @chloealee Can you try doing the string replacement above, before putting it into the CSV parser? I think you might be running into trouble since `$csv` is a CSV reader object down below.

```php
$file = Storage::get($filepath);
$file = str_replace("\r", "\n", $file);

$csv = Reader::createFromString($file);
```

chloealee: omg you are so right, that worked! Thank you @DFurnes !

chloealee:
@ngjo do you think we should test this with Rock the Vote and Facebook Share Excel CSVs?

chloealee:
awesome, thanks @ngjo ! All of those files worked! The only thing is there was an error with the TurboVotes Excel sheet - on line 16, the `refferal-code` is missing the `campaignRunID` value so it errored out: `user:5b5a0fa1567d8d00ad32d62f,campaign:7059,campaignRunID`. But besides that looking like it can handle Excel files now!

@DoSomething/team-bleed this is tested and ready for a 👀 !

PR #46, opened 2018-09-05
Mega files

chloealee:
#### What's this PR do?
Creates new job to create a new turbo vote post in Rogue that is separate from `ImportTurboVotePosts` jobs. 

👀 

Chompy breaks when large files (more than 2k records) are uploaded. @DFurnes suggested that it might be because this job has a lot of logic in it. Lets see if breaking this up into two different jobs helps. 

#### Relevant tickets

Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/159660724) Pivotal Card.


app/Jobs/GetOrCreateUser.php
```diff
@@ -0,0 +1,95 @@
+<?php
```
chloealee: ignore this for now!

app/Jobs/ImportTurboVotePosts.php
```diff
@@ -94,33 +116,11 @@ public function handle(Rogue $rogue)
...
-                            if ($post['data']) {
-                                $this->stats['countPostCreated']++;
-                            }
-                        } catch (\Exception $e) {
-                            info('There was an error storing the post for: ' . $record['id'], [
-                                'Error' => $e->getMessage(),
-                            ]);
-                        }
+                        CreateTurboVotePostInRogue::dispatch($record, $referralCodeValues, $user);
+
```
chloealee: @katiecrane I'm trying to make sure the `$this->stats['countPostCreated']+++` stat increases when a new post is made. I tried to do the `Queue::after` method above (lines 67-76_ but that didn't work/not sure I'm using it correctly. Do you have any ideas on how we can still get this stat count to work?

katiecrane: @chloealee how are you running your queue locally?

chloealee: OHH good call, @katiecrane ! I was doing `sync` but changed to `redis` and ran `php artisan queue:listen` but it timedout :-/ and the file I'm using is a test one with only 18 records 🤔

katiecrane: @chloealee ah I thought an issue could have been using `php artisan queue:work` which I believe only runs what's in the queue when you use the command and doesn't grab subsequent jobs...I ran into an issue with that when working on the queue stuff for longshot

katiecrane: but I've also not had luck with using `queue::after`

chloealee: ah interesting! I tried with `php artisan queue:work` too but the script got `Killed`. I'll keep trying other ways - basically just need to figure out how to tell the first job that the job to create the post in Rogue was successful or not !

katiecrane: @chloealee hmmm interestinggggg I wonder if the second job can edit the stat itself?

chloealee: @katiecrane yes! I originally tried to pass the stat to the job and have it increase but it didn't work the way I tried to do it. Then I thought about it more and realized that it probably actually makes sense for that counter to stay in the original job since it is one data point in the overall stat?

katiecrane: @chloealee were you calling `queue::after` from a service provider?

chloealee: @katiecrane I'm not! I saw that but tried to do from this job (lines above) instead. Maybe that is why it isn't working tho! If I call it from the [service provider](https://github.com/DoSomething/chompy/blob/master/app/Providers/AppServiceProvider.php) though, is that strange for it to increase the post created counter from there?

katiecrane: @chloealee I'm not sure if it could update the stat from there, unless it somehow knew which stat it was?

chloealee: @katiecrane hm yeah that is what I'm thinking too... I'll try to dig around for something besides `queue::after` bc I'm not sure if this route will work!

app/Jobs/ImportTurboVotePosts.php
```diff
@@ -94,33 +100,14 @@ public function handle(Rogue $rogue)
...
-
-                            if ($post['data']) {
-                                $this->stats['countPostCreated']++;
-                            }
-                        } catch (\Exception $e) {
-                            info('There was an error storing the post for: ' . $record['id'], [
-                                'Error' => $e->getMessage(),
-                            ]);
+                        if (CreateTurboVotePostInRogue::dispatch($record, $referralCodeValues, $user)) {
+                            $this->stats['countPostCreated']++;
```
chloealee: @katiecrane I got this to work like this!

chloealee:
@katiecrane I would like to deploy this branch to thor to see if the changes made so far help with uploading larger files. Is this ok with you?

katiecrane:
@chloealee sure! the large file problem - did that not exist locally and we were just running into it on QA/prod?

chloealee:
cool, thanks @katiecrane ! I don't think we ever tried the large files locally so @ngjo saw this when she was trying larger files (@ngjo correct me if I'm wrong!).

chloealee:
@katiecrane it seems to time out and just shows the message `Killed` in the logs but want to try to see how it works on thor!

chloealee:
@katiecrane update - this failed in QA bc of this error:

```
Sep 07 11:56:57 dosomething-chompy-qa app/queue.1: #30 {main}, 2018-09-07 15:56:57)) at /app/vendor/laravel/framework/src/Illuminate/Database/Connection.php:664, Doctrine\\DBAL\\Driver\\PDOException(code: 42S02): SQLSTATE[42S02]: Base table or view not found: 1146 Table 'heroku_851fadb9fd14bd7.failed_jobs' doesn't exist at /app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:82, PDOException(code: 42S02): SQLSTATE[42S02]: Base table or view not found: 1146 Table 'heroku_851fadb9fd14bd7.failed_jobs' doesn't exist at /app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:80) 
```

This seems weird to me bc it seems to be looking for a `failed_jobs` table in Chompy when this sounds like the table in Rogue?

katiecrane:
@chloealee seems like it's trying to put something into the `failed_jobs` table in Chompy which might not exist? Not sure about the Rogue part?

chloealee:
@katiecrane yes that is what I meant that it doesn't exist! Trying to find where this is happening ... I know there is a `failed_jobs` in Rogue so maybe something is being mixed up... stay tuned!

katiecrane:
@chloealee ahhhhhhh sounds like a job in Chompy might be failing but I wonder why?

app/Jobs/ImportTurboVotePosts.php
```diff
@@ -94,32 +97,10 @@ public function handle(Rogue $rogue)
...
-                            if ($post['data']) {
-                                $this->stats['countPostCreated']++;
-                            }
-                        } catch (\Exception $e) {
-                            info('There was an error storing the post for: ' . $record['id'], [
-                                'Error' => $e->getMessage(),
-                            ]);
+                        $post = CreateTurboVotePostInRogue::dispatch($record, $referralCodeValues, $user);
+
+                        if ($post) {
```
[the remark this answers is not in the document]
DFurnes: I don't think that's what's happening here. The `dispatch` method allows chaining additional configuration methods, like `CreateTurboVotePostInRogue::dispatch(…)->onQueue('queueName')` by returning a [`PendingDispatch` object](https://github.com/laravel/framework/blob/aac6144db22768c780ab5a19c1441830254c6057/src/Illuminate/Foundation/Bus/Dispatchable.php#L9-L17) which since it's not `null`, will evaluate truthily in your conditional below.

Because we're using a queue job here, we won't be able to see the result of the job synchronously and will have to find another way to tell an admin that the whole set of jobs has completed. (And that's kind of a good thing, since it means this job can "finish" a lot quicker and is less likely to time out! 😄)

app/Jobs/ImportTurboVotePosts.php
```diff
@@ -66,8 +65,12 @@ public function tags()
     public function handle(Rogue $rogue)
     {
         // @TODO: We need to write some tests for this import!
+        info("getting records");
+
         $records = $this->getCSVRecords($this->filepath);
 
+        info("records received");
```
DFurnes: This reminds me that we need to set up StyleCI on this project to catch nitpicky style problems (these two `info()` calls should be using single-quoted strings according to our style guide). Let's make a ticket so we don't forget!

DFurnes:
> `Sep 07 11:56:57 dosomething-chompy-qa app/queue.1: #30 {main}, 2018-09-07 15:56:57)) at /app/vendor/laravel/framework/src/Illuminate/Database/Connection.php:664, Doctrine\\DBAL\\Driver\\PDOException(code: 42S02): SQLSTATE[42S02]: Base table or view not found: 1146 Table 'heroku_851fadb9fd14bd7.failed_jobs' doesn't exist at /app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:82, PDOException(code: 42S02): SQLSTATE[42S02]: Base table or view not found: 1146 Table 'heroku_851fadb9fd14bd7.failed_jobs' doesn't exist at /app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:80)`
>
> This seems weird to me bc it seems to be looking for a `failed_jobs` table in Chompy when this sounds like the table in Rogue?

Ah yep, we need to create the [`failed_jobs` table](https://laravel.com/docs/5.6/queues#dealing-with-failed-jobs) here too! That's what Laravel uses to keep track of jobs that have failed more than the maximum allowed tries so they can be re-run later. You can generate that migration with Artisan's `queue:failed-table` command.

app/Jobs/ImportTurboVotePosts.php
chloealee: ah yes! [Here's](https://www.pivotaltracker.com/n/projects/2019429/stories/160462809) the card.

app/Jobs/ImportTurboVotePosts.php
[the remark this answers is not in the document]
DFurnes: Yes, this looks great! What we're doing here is splitting the work into the smallest "steps" we can do at once (so first parse through all the rows in the CSV, then one-by-one make the API calls creating each user/post). 

This should help keep things from timing out since the slow part of this whole process is spending X seconds per row to make the API calls to Northstar & Rogue. So for a CSV with 1000 rows: instead of, say, a 1005 second job to read the CSV and process all the rows, we've got one 5 second job to parse the CSV and 1000 one-second jobs to send each row to our backends.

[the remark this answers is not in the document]
DFurnes: For now, let's just remove the stat-tracking in this job since it isn't particularly helpful when this is split into separate jobs (and we can see the same information with a little more work by peeking at the logs for this application).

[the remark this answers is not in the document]
chloealee: ah ok that makes more sense when thinking about not using the stat, thanks @DFurnes ! I was initially thinking if the first job is dependent on the responses from the smaller jobs, then the total time spent in the first job isn't really cut down any since it still waits for the responses of the other jobs. This makes more sense, especially if we cut the stat-tracking!

chloealee:
@DFurnes I'm not sure I understand ! I thought that this was already happening into the individual row's job since it happens below [this](https://github.com/DoSomething/chompy/pull/46/files#diff-ed45aaf47afdecacbc92e2d0393dc00bR75)?

DFurnes:
Anything within the `ImportTurboVotePosts ` job will happen within that job. So while we're moving the "create post" step into a separate `CreateTurboVotePostInRogue` job, we're still making 2-3 API calls per row inside the initial job (get/create user, get post, and optionally update post below). These calls add up (250-1500ms each!), and so that's going to get expensive if we have larger CSVs.

```php
foreach ($records as $offset => $record) {
    // ...
    try {
        $user = $this->getOrCreateUser($record, $referralCodeValues);  // <--- API call!
    } 
    // ...
    if ($user) {
        $post = $rogue->getPost([                                      // <--- API call!
            'campaign_id' => (int) $referralCodeValues['campaign_id'],
            'northstar_id' => $user->id,
            'type' => 'voter-reg',
        ]);
        if (! $post['data']) {
            CreateTurboVotePostInRogue::dispatch($record, $referralCodeValues, $user);
        } else {
            // ...
            try {
                $rogue->updatePost(…);                                 // <--- API call!
```

Ideally the `ImportTurboVotePosts` should _just_ be responsible for iterating over the CSV and kicking off individual jobs per row. Each of those downstream jobs can then make any API requests required to create the user or post for that row.

chloealee:
@DFurnes ah I see what you're saying now! Yes! So I [commented above](https://github.com/DoSomething/chompy/pull/46#issuecomment-421044360) about this but basically, I was under the impression that because we're using queued jobs, a job couldn't return information immediately in which we would need here? e.g. we'd need to get user in order to look up the post and we'd need to know if the post existed or not in order to figure out if we should create or update the post in Rogue. 

For updating the post, I figured since it was only one line basically, that creating a new job wouldn't be that big of a difference but I can certainly do that !

DFurnes:
@chloealee Yep, each of those don't need to be individual jobs (because, right, they're dependent on each other). We shouldn't, however, have expensive API requests in the main job so we want everything inside that `foreach` loop to be moved into the per-row job.

chloealee:
@DFurnes oohhhh you mean to move all that logic into the `CreateTurboVotePostInRogue` job?

DFurnes:
Right!

chloealee:
ah I see what you mean! Super smart - thanks for the explanation on this and for your patience! :)

chloealee:
thanks @DFurnes ! Right now, I'm working on actually sending over all of the `$records` to the job instead of one by one so we can still count all the stats we want! I've been struggling with `Exception
Serialization of 'Closure' is not allowed` but working on that and will let you know once I've figured it out!

Before I go down a rabbithole though, does that sound ok to you to send all of the `$records` over to the `CreateTurboVotePostsInRogue` job and then do the iteration there?

DFurnes:
Then we'd be right back to square zero!! We want to split up the work into multiple jobs so that we don't run into timeout issues (and also so that we can have more discrete points to "retry" work from if we run into an issue). For tracking progress, I think we want to think about a better way to surface how many things are waiting in the queue for users (or broadcast progress events to a user-facing log).

chloealee:
@DFurnes ah duh! Ok I will stop that then! So I guess we just lose all of the stats for now. Should I try this on QA ?

DFurnes:
Yeah, let's see how this works on QA and we can work on improving the admin experience afterwards.

chloealee:
@DFurnes this is still running (yay!), which is great news! The only thing that isn't working perfectly is the progress log. It seems to be stuck here:

screen shot 2018-09-14 at 4 43 19 pm

DFurnes:
That's great! Yep, I think that sounds like a perfect plan.

chloealee:
awesome, thanks @DFurnes (and for all the explanation & help!)!

PR #47, opened 2018-09-20
Separates Rock The Vote Import into two jobs to enable larger file imports

chloealee:
#### What's this PR do?
Separates Rock The Vote Import into two jobs to enable larger file imports.
 - One job to process CSV.
 - One job to create post in Rogue.

- Does all the logic seem ok? 

- Unfortunately, we will have to get rid of the stats when we separate these jobs. Are we ok with this for now? We should still have all the info. in the logs though! 

#### Relevant tickets

Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/159660724) Pivotal Card.


app/Jobs/CreateTurboVotePostInRogue.php
```diff
@@ -55,7 +54,7 @@ public function handle(Rogue $rogue)
                 ]);
             }
 
-            if ($user) {
+            if (isset($user)) {
```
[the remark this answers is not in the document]
katiecrane: good catch! thanks for explaining :)

app/Jobs/ImportRockTheVotePosts.php
```diff
@@ -26,12 +26,10 @@ class ImportRockTheVotePosts implements ShouldQueue
     protected $filepath;
 
     /**
-     * The path to the stored csv.
+     * The total records in the stored csv.
```
katiecrane: is this a count of the total records or is it the records themselves?

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -0,0 +1,392 @@
...
+                        'source' => 'rock-the-vote',
+                        'source_details' => $sourceDetails,
+                        'details' => $postDetails,
+                    ];
+
+                    try {
+                        $post = $rogue->createPost($postData);
+
+                        if ($post['data']) {
+                            info('post created in rogue for ' . $this->record['id']);
```
katiecrane: there's no "id" column in RTV csvs, I believe previously we were using email address?

katiecrane:
@chloealee it's really hard to tell if any of the logic changed in the fresh RTV file, I'm trying to figure out a good way to compare it to the old logic - do you know if you changed anything after you copied and pasted?

app/Jobs/ImportRockTheVotePosts.php
chloealee: @katiecrane great q! This is [the count](https://github.com/DoSomething/chompy/blob/7b32036b779ce702f50076bc1e60c8c87c75554f/app/Traits/ImportToRogue.php#L35) of the total records. Updated this comment to make it more clear!

app/Jobs/CreateRockTheVotePostInRogue.php
chloealee: ah great catch! Updated :)

chloealee:
@katiecrane ah yes it is hard to tell ! You could maybe compare to [this file](https://github.com/DoSomething/chompy/blob/master/app/Jobs/ImportRockTheVotePosts.php)?

chloealee:
@katiecrane the only thing that I added was that log line that you caught before that had `$record['id']` instead of `$record['Email address']` and deleted the stats stuff since it won't work to have in separate two jobs!

katiecrane:
@chloealee okay I think we're good then! Let's just make sure to do some spot checking on QA just in case! great work!

chloealee:
yes definitely @katiecrane ! :)

PR #48, opened 2018-09-21
RTV: Make sure campaign ids are integers before using

katiecrane:
#### What's this PR do?
- Adds an additional check before we use the given campaign id and run id. We used to just use them if both existed, now we only use them if both exist AND ARE INTEGERS.

Does it do what I think it does?

We get some wild tracking sources, but we don't want that to break anything.

#### Relevant tickets

[Card](https://www.pivotaltracker.com/story/show/160321941)

### DIFF
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -192,8 +192,8 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
 
         // See if we have all the required information we need
         if (!array_has($values, ['northstar_id', 'campaign_id', 'campaign_run_id'])) {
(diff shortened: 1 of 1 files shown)


chloealee:
Ah good catch, @katiecrane ! One question (and I'm not sure if this matters) - I think all Contentful `campaign_id`s are strings? Will this break anything in the future if we upload posts from Contentful campaigns?

katiecrane:
oh poop. great question - I'm going to ask @ngjo in #dev-chompy

katiecrane:
@chloealee: @DFurnes says no plans to use Contentful ids outside Contentful! https://dosomething.slack.com/archives/CA0N5FXND/p1537556275000100

chloealee:
awesome, thanks for closing the loop on that @katiecrane !

PR #49, opened 2018-09-21
increases upload_max-filesize and adds post_max_size

chloealee:
#### What's this PR do?
When uploading a large RTV file, Chompy breaks - it tries to upload the large file and in the bottom left corner of my browser says “Uploading 5%” and then keeps restarting and then breaks with this image: 
screen shot 2018-09-21 at 1 49 35 pm

In the logs, I see `PHP Warning:  POST Content-Length of 20804247 bytes exceeds the limit of 8388608 bytes in Unknown on line 0` and `"Server Request Interrupted" method=POST path="/import" host=importer-qa.dosomething.org request_id=5af63bcd-4c70-45dd-bcaa-962febf1a4ba fwd="73.61.23.2" dyno=web.1 connect=0ms service=301ms status=503 bytes=21023 protocol=https `. This PR increases `upload_max-filesize` and adds `post_max_size` to hopefully fix this error. 

👀 

#### Relevant tickets
Related to work done in this PR: #47

### DIFF
--- a/public/.user.ini
+++ b/public/.user.ini
@@ -1,2 +1,3 @@
-upload_max_filesize = 10M
+upload_max_filesize = 1000M
 memory_limit = 256M
+post_max_size = 1000M
--- a/public/.user.ini
+++ b/public/.user.ini
@@ -1,3 +1,3 @@
-upload_max_filesize = 2000M
-memory_limit = 2048M
-post_max_size = 2000M
+upload_max_filesize = 500M
+memory_limit = 512M
+post_max_size = 500M
--- a/public/.user.ini
+++ b/public/.user.ini
@@ -1,3 +1,3 @@
-upload_max_filesize = 1000M
+upload_max_filesize = 2000M
 memory_limit = 256M
-post_max_size = 1000M
+post_max_size = 2000M
--- a/public/.user.ini
+++ b/public/.user.ini
@@ -1,3 +1,3 @@
 upload_max_filesize = 2000M
-memory_limit = 256M
+memory_limit = 2048M
 post_max_size = 2000M


chloealee:
blah - this seemed to _help_ but still broke! It made it to 100% of the uploading stage but when it was then waiting for a response from Chompy, it broke :( 

Here are some things it said in the logs:
`
Sep 24 14:25:41 dosomething-chompy-qa app/heroku-redis: source=REDIS addon=redis-rectangular-74179 sample#active-connections=2 sample#load-avg-1m=0.065 sample#load-avg-5m=0.09 sample#load-avg-15m=0.09 sample#read-iops=0 sample#write-iops=0 sample#memory-total=15664360kB sample#memory-free=11355520kB sample#memory-cached=1351372kB sample#memory-redis=323736bytes sample#hit-rate=0.28827 sample#evicted-keys=0 
`

`
Sep 24 14:27:32 dosomething-chompy-qa app/web.1: [24-Sep-2018 18:27:32 UTC] PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 134217736 bytes) in /app/vendor/league/csv/src/Stream.php on line 396 
`

`
Sep 24 14:27:32 dosomething-chompy-qa app/web.1: [24-Sep-2018 18:27:32 UTC] [2018-09-24 18:27:32] production.ERROR: Allowed memory size of 268435456 bytes exhausted (tried to allocate 134217736 bytes) {"userId":31,"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalErrorException(code: 1): Allowed memory size of 268435456 bytes exhausted (tried to allocate 134217736 bytes) at /app/vendor/league/csv/src/Stream.php:396) 
`

It seems like it is still using too much memory 🤔 

@DoSomething/team-bleed I _think_ this is still just happening at the stage when it is reading the CSV. Since we've already broken the job into two steps, with the first being the job to read the CSV, I don't think there is anything else we can do code level wise and might have to either: 
- Increase memory somewhere else. 
- Still break CSVs into smaller units and find out the maximum records we can upload (which still will be better than what Jen was doing before at 1-2k records). 

@DoSomething/team-bleed any other thoughts/solutions I'm missing?

chloealee:
ooo after increasing the `memory_limit`, things seem to start! More updates to come...

ngjo:
My two cents -- if we can get through ~10k that's a significant improvement to files of ~1200 and I'm not sure if it's worth continuing to dig. The first upload will be a 100k (which we can break up) but then subsequent uploads will be more in the 10k range!

chloealee:
@ngjo ok, thanks for your thoughts on this! I'll take the last 15k of the csv now and test uploading. Stay tuned!

chloealee:
ok so this just stopped again at the [9,415th record](https://papertrailapp.com/systems/dosomething-chompy-qa/events?focus=981358145396543490&selected=981358145396543490) - doesn't _quite_ get to an even number but I think it is safe to upload 9k records at a time if that works for you @ngjo ? 

From the 9,415 records that it went through, 8,360 posts were created.  This job took about 2 hours and 18 min to get through those records and create posts.

ngjo:
Great, thanks @chloealee ! I'll have the file broken up into 9k chunks. When I import it, should I rely on the progress bar to know that it got through or should I notify you when I upload to monitor on the back?

chloealee:
@ngjo as along as nothing breaks, the progress bar should be an accurate representation! Can you also notify me when you upload though so I can monitor for safe measure?

chloealee:
@DoSomething/team-bleed if we're fine with uploading 9k records, this is ready for review!

public/.user.ini
```diff
@@ -1,2 +1,3 @@
-upload_max_filesize = 10M
-memory_limit = 256M
\ No newline at end of file
+upload_max_filesize = 2000M
+memory_limit = 2048M
```
DFurnes: Have you been upping the dyno size when testing this setting? Chompy QA is running on a hobby dyno, which [has only 512MB of RAM](https://devcenter.heroku.com/articles/dyno-types).

chloealee: omg I didn't even check that! I didn't up the dyno size - should I do that to standard-2-x?

chloealee: @DFurnes what is the difference between upping this in the code versus on Heroku?

DFurnes: > omg I didn't even check that! I didn't up the dyno size - should I do that to `standard-2-x`?

I think that's a good first try. We probably don't want to be paying for performance dynos for this application, so if that doesn't help then we'll have to find solutions elsewhere.

> @DFurnes what is the difference between upping this in the code versus on Heroku?

You're telling PHP how much memory it is _allowed_ to use per-request when setting [`memory_limit`](http://php.net/manual/en/ini.core.php#ini.memory-limit), but there's still a "cap" on how much memory is available (and even if you ask nicely, you can't use more than you have).

DFurnes: Yeah, let's give it a try at least!

chloealee: @DFurnes ok! So this time after increasing the dyno size, it got through 54,877 records! What I would love to do is:
- Wipe NS QA tonight/tomorrow am before office hours (optional, I can also just try last 50k records even though there might be some overlap)
- Re-test with 50k records

If all goes well tomorrow, we can keep that dyno for a bit and merge this to master? 
cc @DoSomething/team-bleed

DFurnes: Woohoo, that's much nicer! 👏 Totally onboard for a QA reset, just drop a note in #deploys!

chloealee: cool, will do!! :)

chloealee: Just want to give an update since I know @ngjo wants to do some RTV chomping! This has been running since 10:40am. It's only created 4,878 new posts and has only gone through 3,273 records it looks like... I'm not sure why it is so slow. The thor refresh did get kicked off after I started the script started so not sure if that affected anything?

katiecrane: @chloealee does clearing the cache work for that? I'm not sure. Also the Northstar QA database has been struggling all day so I think the errors you're seeing are due to that

chloealee: ah thanks for letting me know about that @katiecrane ! I think that _might_ be causing the job to run so slow. Perhaps I'll just keep it going then and continue to keep an eye on it rather than starting from scratch since that likely won't change any behavior if NS QA has been struggling all day. Thanks, Katie!

chloealee: stopped this job (by clearing the cache - thanks @katiecrane !) as it was craaawling and I think we can maybe regroup as a team re: next steps on Monday since some people are out and we probably wouldn't get to RTV imports then anyway! 

At first, It seemed like this job was really heavy for Northstar QA and slowed it down a lot. @katiecrane made a great point that even though we beefed up the queue dyno for Chompy QA, it might have been too much for Northstar QA. 

In more than 24 hours, it seemed to only get through getting the user info. of ~8,800 records.

After stopping this job, Northstar QA's performance didn't improve so perhaps this was just a coincidence and that is why the RTV import was moving so slowly. 

Lets re-group on Monday re: next steps and by then we should know more info. about NS too!

chloealee: hi team! I would like to re-start this on QA and test again but wanted to check here first - is this ok with everyone?

chloealee: @DoSomething/team-bleed looks like this stopped after creating 8,155 new voter reg posts of the 50k record import :-/ It was running for almost exactly 4 hours when it recorded the last thing of from this job in the [logs](https://papertrailapp.com/systems/dosomething-chompy-qa/events?focus=986764510621347841&selected=986764510621347841). I don't see anything in the logs to indicate any errors though. @DFurnes is there a way to see how much memory we used of the 1 GB of RAM in heroku?

DFurnes: @chloealee You can look at metrics per formation in [Heroku's metrics tab](https://dashboard.heroku.com/apps/dosomething-chompy-qa/metrics/queue). It looks like we never used above ~200MB of RAM (which makes sense to me now that we've split these jobs out)!

It might be helpful to jump on a call tomorrow or Friday and talk this through to see if there's anything we're missing that could be causing trouble here, but IMO this PR is totally good to merge and deploy.

chloealee: cool - thank you for sharing that @DFurnes ! 

I'd love to jump on a call today if you're free - I don't have any meetings at all today so just lmk when you're free! :) Would love to discuss if there is anything we're missing and if we can't think of anything, what is a safe number of records per file we can tell @ngjo is safe to upload and that we're confident will make through the entire file.

chloealee:
@DFurnes updated the settings!

PR #50, opened 2018-09-26
Send voter reg status to Northstar

katiecrane:
#### What's this PR do?
- Added a `updateNorthstarStatus` function to each voter reg import. I know before we've had stuff as traits but I think that makes it a little harder for debugging. Open to suggestions!
- Call this function after we get or create a user.

If you run an import, do these users have `voter_registration_status` set in Northstar?

Before merge: Update Northstar to have a `voter_registration_status` setter that only allows this field to move up the hierarchy.

#### Relevant tickets

[Card](https://www.pivotaltracker.com/story/show/160103784)

### DIFF
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -55,6 +55,7 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
             }
 
             if (isset($user)) {
+                // @TODO: If we write more functions that are identical across all voter reg imports, pull out into a ImportsVoterReg trait
                 $this->updateNorthstarStatus($user, $this->translateStatus($this->record['Status'], $this->record['Finish with State']));
 
                 $post = $rogue->getPost([
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -55,6 +55,8 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
             }
 
             if (isset($user)) {
+                $this->updateNorthstarStatus($user, $this->translateStatus($this->record['Status'], $this->record['Finish with State']));
+
                 $post = $rogue->getPost([
                     'campaign_id' => (int) $referralCodeValues['campaign_id'],
                     'northstar_id' => $user->id,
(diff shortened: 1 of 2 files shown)


app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -55,6 +55,8 @@ public function handle(Rogue $rogue)
             }
 
             if (isset($user)) {
+                $this->updateNorthstarStatus($user, $this->translateStatus($this->record['Status'], $this->record['Finish with State']));
+
```
chloealee: @katiecrane I could be totally missing something but I think `updateNorthstarStatus` only takes two params?

chloealee:
> I know before we've had stuff as traits but I think that makes it a little harder for debugging. Open to suggestions!

Totally hear you on this! The only hesitation I have about not using traits is that it seems a little confusing and inconsistent to use traits sometimes and other times to not. So I guess maybe we should just pick which way we want to do and just do that?

app/Jobs/CreateRockTheVotePostInRogue.php
katiecrane: @chloealee this is two params! `$user` and `$this->translateStatus($this->record['Status'], $this->record['Finish with State'])`

chloealee: omg my old eyes!

chloealee:
@katiecrane that totally makes sense to me too! I don't have terribly strong feelings on this but if we want to stay consistent and break out the same functions into the traits, maybe we can just add more comments to make it more explicit that we use the function in the trait/or where we don't and hopefully easier to debug?

chloealee:
@katiecrane oooo good call! I didn't realize that it would be in a different trait and I think you're right that that makes sense! Since this would be the only thing in the new trait, maybe it _is_ ok to just leave here but maybe make an `@TODO` comment if there seems to be more overlapping voter reg specific functions to make a new trait? Sorry I keep going back and forth on this!

katiecrane:
@chloealee oooo I like the idea of the `@TODO` for now since it would just be one thing in the trait which feels a lil silly. I've been waffling too! I feel like I do have an opinion but where we land on this is _that_ big of a deal

chloealee:
haha agreed, @katiecrane ! If it feels good to you, I say we just do the `@TODO` for now!

Issue #51, opened 2018-10-09
Investigation: Errors when importing Rock The Vote posts.


DFurnes:
## _BUG_

Jen has seen some _suspicious_ behavior when uploading the past few Rock The Vote CSVs. We want to track down the source of the error and make sure that everything is safely making it in to our backend.

DFurnes:
We've had 10 failed runs of the `ImportRockTheVotePosts` job, likely from multiple attempts on a few CSVs. Here's the list of failed jobs:

```
Running php artisan queue:failed on ⬢ dosomething-chompy... up, run.9459 (Standard-1X)
+----+------------+----------+------------------------------------+---------------------+
| ID | Connection | Queue    | Class                              | Failed At           |
+----+------------+----------+------------------------------------+---------------------+
| 91 | redis      | importer | Chompy\Jobs\ImportRockTheVotePosts | 2018-10-09 16:07:21 |
| 81 | redis      | importer | Chompy\Jobs\ImportRockTheVotePosts | 2018-10-09 15:37:27 |
| 71 | redis      | importer | Chompy\Jobs\ImportRockTheVotePosts | 2018-10-09 15:11:46 |
| 61 | redis      | importer | Chompy\Jobs\ImportRockTheVotePosts | 2018-10-09 14:40:58 |
| 51 | redis      | importer | Chompy\Jobs\ImportRockTheVotePosts | 2018-10-05 21:02:03 |
| 41 | redis      | importer | Chompy\Jobs\ImportRockTheVotePosts | 2018-10-05 20:41:27 |
| 31 | redis      | importer | Chompy\Jobs\ImportRockTheVotePosts | 2018-10-05 20:16:00 |
| 21 | redis      | importer | Chompy\Jobs\ImportRockTheVotePosts | 2018-10-05 18:19:17 |
| 11 | redis      | importer | Chompy\Jobs\ImportRockTheVotePosts | 2018-10-05 18:10:51 |
| 1  | redis      | importer | Chompy\Jobs\ImportRockTheVotePosts | 2018-10-05 17:38:06 |
+----+------------+----------+------------------------------------+---------------------+
```

DFurnes:
All of the jobs failed with the same issue, encoding a row for the `CreateRockTheVotePostInRogue` job:

```
Illuminate\Queue\InvalidPayloadException: Unable to JSON encode payload. Error code: 5 in /app/vendor/laravel/framework/src/Illuminate/Queue/Queue.php:83
Stack trace:
[...]

PR #53, opened 2018-10-16
Fail job when there are errors instead of just logging and continuing 

chloealee:
#### What's this PR do?

Throws an exception for any jobs that are not able to be successfully completed instead of just logging and continuing. 

Does the job fail and enter the failed jobs table? 

#### Relevant tickets

Fixes [this issue](https://github.com/DoSomething/chompy/issues/51) and [this Pivotal card](https://www.pivotaltracker.com/n/projects/2019429/stories/161184640).

### DIFF
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -2,6 +2,7 @@
 
 namespace Chompy\Jobs;
 
+use Exception;
 use Carbon\Carbon;
 use Chompy\Services\Rogue;
 use Illuminate\Bus\Queueable;
@@ -9,9 +10,8 @@ use Chompy\Traits\ImportToRogue;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Foundation\Bus\Dispatchable;
-use Symfony\Component\HttpKernel\Exception\HttpException;
-
 class CreateRockTheVotePostInRogue implements ShouldQueue
+
 {
     use Dispatchable, InteractsWithQueue, Queueable, ImportToRogue;
 
@@ -264,7 +264,7 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
             if ($user->id) {
                 info('created user', ['user' => $user->id]);
             } else {
-                throw new HttpException(500, 'Unable to create user: ' . $record['Email address']);
+                throw new Exception(500, 'Unable to create user: ' . $record['Email address']);
             }
         }
 
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -75,17 +75,8 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
                     'details' => $postDetails,
                 ];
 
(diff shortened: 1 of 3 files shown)


app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -285,6 +263,8 @@ private function getOrCreateUser($record, $values)
 
             if ($user->id) {
                 info('created user', ['user' => $user->id]);
+            } else {
+                throw new HttpException(500, 'Unable to create user: ' . $record['Email address']);
```
DFurnes: Instead of an `HttpException`, let's just throw a normal `\Exception`. HttpExceptions are automatically caught by Laravel to render as a nice error page, but since this isn't happening in a web process that doesn't do much for us.

app/Services/Rogue.php
```diff
@@ -72,6 +74,10 @@ public function createPost($data)
 
         $post = $this->asClient()->send('POST', 'v3/posts', ['multipart' => $multipartData]);
 
+        if (! $post['data']) {
+            throw new HttpException(500, 'Unable to create post for user: ' . $data['northstar_id']);
```
DFurnes: Same here!

DFurnes:
Yeah, that sounds like a good idea to me!

chloealee:
@DFurnes this is so strange! I changed `NORTHSTAR_AUTHORIZATION_ID=d` and even did `NORTHSTAR_AUTHORIZATION_SECRET=gibberish` and `NORTHSTAR_CLIENT_SECRET=gibberish` but the posts were still able to be created 🤔

DFurnes:
@chloealee We keep an access token in the `clients` table so that we don't need to fetch a new one from Northstar every time we want to make a request, so you'll need to clear that too. (And then it'll use `NORTHSTAR_CLIENT_ID` and `NORTHSTAR_CLIENT_SECRET` to fetch a new one.)

chloealee:
@DFurnes a-ha! Ok so that worked but it hits this page and nothing is in the `failed_jobs` table :-/ 
screen shot 2018-10-17 at 3 50 56 pm

DFurnes:
@chloealee What queue driver are you using on your local?

chloealee:
@DFurnes `sync`

DFurnes:
Ah yep! So the `sync` driver runs jobs immediately within a request (rather than queueing them up in Redis, SQS, or whatever). It's useful for development since you don't need to remember to hit `php artisan queue:work` in another tab, but I also don't think it supports retries or failing jobs.

Can you try swapping `QUEUE_DRIVER` to `redis` and testing with that?

chloealee:
oh interesting! I changed to `redis` and it looks like it is just getting stuck in a loop going through the users, getting stuck at `getting user with the email field` line but not failing the job

chloealee:
Merging this PR, per @DFurnes 's [Slack approval](https://dosomething.slack.com/archives/CA0N5FXND/p1539870378000100?thread_ts=1539792868.000100&cid=CA0N5FXND)!

DFurnes:
Woo, yeah I think this looks good!

> I wonder if it is because the error isn't handled here?

We want to leave errors like that unhandled!! That's how we know to fail a job.

chloealee:
oooh makes sense!

PR #58, opened 2019-02-08
Adds new user email_subscription_status to RTV import

aaronschachter:
#### What's this PR do?

Adds logic to the RTV import to set the `email_subscription_status` field for new users based on the `Opt-in to Partner email?` column in a given csv.

Ensure Rogue DB has correct data for import. For our test file, we need a campaign with id 8017 (per hardcoded ID) and an action with properties:

* `name`: `'january-2019-rockthevote'`
* `campaign_id`: 8017
* `post_type`: `'voter-reg'`

Example:
```
INSERT INTO `actions` (`id`, `name`, `campaign_id`, `post_type`, `reportback`, `civic_action`, `scholarship_entry`, `active`, `noun`, `verb`, `created_at`, `updated_at`, `deleted_at`)
VALUES
	(1, 'january-2019-rockthevote', 8017, 'voter-reg', 1, 1, 1, 0, 'things', 'done', '2019-02-08 15:01:20', '2019-02-08 15:01:20', NULL);

```

This file could use some more TLC in a future PR (e.g. removing sending a Campaign Run ID)

#### Relevant tickets

Fixes https://www.pivotaltracker.com/story/show/163564896

### DIFF
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -253,11 +253,15 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
                 'addr_state' => $record['Home state'],
                 'addr_zip' => $record['Home zip code'],
                 'source' => env('NORTHSTAR_CLIENT_ID'),
-                'email_subscription_status' => $this->transformEmailSubscriptionStatus($record['Opt-in to Partner email?'])
             ];
 
             if ($record['Phone']) {
                 $userData['sms_status'] = $this->transformSmsStatus($record['Opt-in to Partner SMS/robocall']);
+            }
+
+            $emailSubscriptionColumn = $record['Opt-in to Partner email?'];
+            if ($emailSubscriptionColumn) {
+                $userData['email_subscription_status'] = $emailSubscriptionColumn === 'Yes';
             }
 
             $user = gateway('northstar')->asClient()->createUser($userData);
@@ -355,21 +359,6 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
         $indexOfNewStatus = array_search($newStatus, $statusHierarchy);
 
         return $indexOfCurrentStatus < $indexOfNewStatus ? $newStatus : null;
-    }
-
-    /*
-     * Translate "Opt-in to Partner email?" from Rock the Vote CSV to a Northstar email_subscription_status
-     *
-     * @param array $sms_status
-     * @return string
-    */
-    private function transformEmailSubscriptionStatus($email_subscription_status)
-    {
-        if ($email_subscription_status === 'Yes') {
-            return true;
-        }
-
-        return false;
     }
 
     /*
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -253,6 +253,7 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
                 'addr_state' => $record['Home state'],
                 'addr_zip' => $record['Home zip code'],
                 'source' => env('NORTHSTAR_CLIENT_ID'),
+                'email_subscription_status' => $this->transformEmailSubscriptionStatus($record['Opt-in to Partner email?'])
             ];
 
             if ($record['Phone']) {
@@ -356,6 +357,20 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
         return $indexOfCurrentStatus < $indexOfNewStatus ? $newStatus : null;
     }
 
+    /*
+     * Translate "Opt-in to Partner email?" from Rock the Vote CSV to a Northstar email_subscription_status
+     *
+     * @param array $sms_status
+     * @return string
+    */
+    private function transformEmailSubscriptionStatus($email_subscription_status)
+    {
+        if ($email_subscription_status === 'Yes') {
+            return true;
+        }
+
+        return false;
+    }
 
     /*
      * Translate "Opt-in to Partner SMS/robocall" from Rock the Vote CSV to a Northstar sms_status
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -61,13 +61,13 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
 
         $campaignId = (int) $referralCodeValues['campaign_id'];
         $postType = 'voter-reg';
-        // Check if post exists.
+
         $existingPosts = $rogue->getPost([
             'campaign_id' => $campaignId,
             'northstar_id' => $user->id,
             'type' => $postType,
         ]);
-        // If post does not exist, create it.
+
         if (!$existingPosts['data']) {
             info('post not found for user ' . $user->id);
             $rtvCreatedAtMonth = strtolower(Carbon::parse($this->record['Started registration'])->format('F-Y'));
@@ -87,16 +87,16 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
 
             $post = $rogue->createPost($postData);
             info('post created in rogue for ' . $this->recordEmail);
-        // Else if post exists, update post status if required.
-        } else {
-            $postId = $existingPosts['data'][0]['id'];
-            info($postType.' post '.$postId.' found for user ' . $user->id.' and campaign '.$campaignId);
-            $newStatus = $this->translateStatus($this->record['Status'], $this->record['Finish with State']);
-            $statusShouldChange = $this->updateStatus($existingPosts['data'][0]['status'], $newStatus);
+            return;
+        }
 
-            if ($statusShouldChange) {
-                $rogue->updatePost($postId, ['status' => $statusShouldChange]);
-            }
+        $postId = $existingPosts['data'][0]['id'];
+        info($postType.' post '.$postId.' found for user ' . $user->id.' and campaign '.$campaignId);
+        $newStatus = $this->translateStatus($this->record['Status'], $this->record['Finish with State']);
+        $statusShouldChange = $this->updateStatus($existingPosts['data'][0]['status'], $newStatus);
+
+        if ($statusShouldChange) {
+            $rogue->updatePost($postId, ['status' => $statusShouldChange]);
         }
     }
 
@@ -286,11 +286,10 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
 
         $user = gateway('northstar')->asClient()->createUser($userData);
 
-        if ($user->id) {
-            info('created user', ['user' => $user->id]);
-        } else {
+        if (!$user->id) {
             throw new Exception(500, 'Unable to create user: ' . $this->recordEmail);
         }
+        info('created user', ['user' => $user->id]);
 
         return $user;
     }
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -11,6 +11,18 @@ use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Foundation\Bus\Dispatchable;
 
+class RockTheVoteRecord {
+    public function __construct($record)
+    {
+        $this->email = $record['Email address'];
+        $this->mobile = $record['Phone'];
+        $this->email_opt_in = $record['Opt-in to Partner email?'];
+        // Note: Not a typo, this column name does not have the trailing question mark.
+        $this->sms_opt_in = $record['Opt-in to Partner SMS/robocall'];
+        // TODO: Add all other properties used for getting/validating/creating users/posts.
+    }
+}
+
 class CreateRockTheVotePostInRogue implements ShouldQueue
 {
     use Dispatchable, InteractsWithQueue, Queueable, ImportToRogue;
@@ -29,9 +41,9 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
      */
     public function __construct($record)
     {
-    	$this->record = $record;
-        $this->recordEmail = $record['Email address'];
-        $this->recordMobile = $record['Phone'];
+    	$this->data = new RockTheVoteRecord($record);
+        // TODO: Remove once all $this->record references have been replaced with $this->data.
+        $this->record = $record;
     }
 
     /**
@@ -42,11 +54,11 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
     public function handle(Rogue $rogue)
     {
     	if ($this->isTestData($this->record)) {
-            info('progress_log: Skipping test: ' . $this->recordEmail);
+            info('progress_log: Skipping test: ' . $this->data->email);
             return;
         }
 
-        info('progress_log: Processing: ' . $this->recordEmail);
+        info('progress_log: Processing: ' . $this->data->email);
 
         $referralCodeValues = $this->parseReferralCode($this->record['Tracking Source']);
         info('Referral code: ' . implode(', ', $referralCodeValues));
@@ -86,7 +98,7 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
             ];
 
             $post = $rogue->createPost($postData);
-            info('post created in rogue for ' . $this->recordEmail);
+            info('post created in rogue for ' . $this->data->email);
             return;
         }
 
@@ -240,16 +252,16 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
                 return $user;
             }
         }
-        if ($this->recordEmail) {
-            $user = gateway('northstar')->asClient()->getUser('email', $this->recordEmail);
+        if ($this->data->email) {
+            $user = gateway('northstar')->asClient()->getUser('email', $this->data->email);
             if ($user && $user->id) {
                 return $user;
             }
         }
-        if (!$this->recordMobile) {
+        if (!$this->data->mobile) {
             return null;
         }
-        return gateway('northstar')->asClient()->getUser('mobile', $this->recordMobile);
+        return gateway('northstar')->asClient()->getUser('mobile', $this->data->mobile);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -253,6 +253,7 @@ private function getOrCreateUser($record, $values)
                 'addr_state' => $record['Home state'],
                 'addr_zip' => $record['Home zip code'],
                 'source' => env('NORTHSTAR_CLIENT_ID'),
+                'email_subscription_status' => $this->transformEmailSubscriptionStatus($record['Opt-in to Partner email?'])
```
aaronschachter: Can we always assume this column value will be sent, or should we conditionally check whether the column exists before setting? cc @ngjo

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -356,6 +357,20 @@ private function updateStatus($currentStatus, $newStatus)
...
 
+    /*
+     * Translate "Opt-in to Partner email?" from Rock the Vote CSV to a Northstar email_subscription_status
+     *
+     * @param array $sms_status
+     * @return string
+    */
+    private function transformEmailSubscriptionStatus($email_subscription_status)
+    {
+        if ($email_subscription_status === 'Yes') {
```
aaronschachter: Similarly -- do we need to check for any potential variations like `yes`, `Y`?

app/Jobs/CreateRockTheVotePostInRogue.php
ngjo: The column *should* always be sent as long as RTV doesn't change their export settings or the name of the column! It hasn't changed in the past year that we've been exporting/ingesting so there's likely only a slim chance that that column wouldn't be there. If it isn't there for whatever reason, would it just break the import or would everyone get subscribed or something?

aaronschachter: As is, everyone would have their `email_subscription_status` set to false (assuming I figure out how to set it to false per question I flagged in the PR description)

app/Jobs/CreateRockTheVotePostInRogue.php
ngjo: It should be `Yes` but if it's low lift to check for at least `yes` that would more likely than it changing to `Y` on us suddenly. But again, yea, dependent on if RTV changes things

app/Jobs/CreateRockTheVotePostInRogue.php
aaronschachter: Updated this so we only update the value if the column exists.

aaronschachter:
I've been thinking it could make sense to programmatically create the actions with the correct name per date if they don't exist instead of relying on an admin to add them through the Rogue UI... (as well as creating the campaign for 8017 if doesn't exist, if working locally with a scratch DB)

DFurnes:
> I've been thinking it could make sense to programmatically create the actions with the correct name per date if they don't exist instead of relying on an admin to add them through the Rogue UI... (as well as creating the campaign for 8017 if doesn't exist, if working locally with a scratch DB)

I think I'd recommend against this, since it could lead to us accidentally creating campaigns or actions based on a "bad" CSV & not realizing that it's happened. If Chompy runs into trouble like this when creating a post, it should mark that job as "failed" and we can always use that as an opportunity to create the missing campaign/action and re-run with `php artisan queue:retry all`.

app/Jobs/CreateRockTheVotePostInRogue.php
DFurnes: Yeah, I feel like a `str_to_boolean` helper could be a nice safeguard here. 👍

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -210,6 +210,23 @@ private function parseReferralCode($referralCode)
         return $finalValues;
     }
 
+    /**
+     * Parse a CSV field value as boolean.
+     *
+     * @param string $value
+     * @return boolean
+     */
+    private function parseBoolean($value) {
```
aaronschachter: @DoSomething/team-bleed Is there a better place to add this function, to potentially re-use in other imports?

katiecrane: @aaronschachter we don't currently have a helper file because I think so far everything has been specific to a particular import, but this could be a good time to make one if you want to ☕️

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -222,11 +222,14 @@ private function parseReferralCode($referralCode)
...
+        $recordMobile = $record['Phone'];
 
         $userFieldsToLookFor = [
-            'id' => isset($values['northstar_id']) && !empty($values['northstar_id']) ? $values['northstar_id'] : null,
-            'email' => isset($record['Email address']) && !empty($record['Email address']) ? $record['Email address'] : null,
-            'mobile' => isset($record['Phone']) && !empty($record['Phone']) ? $record['Phone'] : null,
+            'id' => isset($userIdValue) && !empty($userIdValue) ? $userIdValue : null,
+            'email' => isset($recordEmail) && !empty($recordEmail) ? $recordEmail : null,
+            'mobile' => isset($recordMobile) && !empty($recordMobile) ? $recordMobile : null,
         ];
```
chloealee: so much easier to read!

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -255,16 +258,23 @@ private function getOrCreateUser($record, $values)
                 'source' => env('NORTHSTAR_CLIENT_ID'),
             ];
 
-            if ($record['Phone']) {
-                $userData['sms_status'] = $this->transformSmsStatus($record['Opt-in to Partner SMS/robocall']);
+            $recordEmailOptIn = $record['Opt-in to Partner email?'];
+            if ($recordEmailOptIn) {
```
chloealee: do we need to check if this `isset` and `!empty`?

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -255,16 +258,23 @@ private function getOrCreateUser($record, $values)
...
+                $userData['email_subscription_status'] = str_to_boolean($recordEmailOptIn);
+            }
+
+            // Note: Not a typo -- this column name does not have the trailing question mark.
+            $recordSmsOptIn = $record['Opt-in to Partner SMS/robocall'];
+            if ($recordSmsOptIn && $recordMobile) {
```
chloealee: same here - do we need to check if `isset($recordSmsOptIn) && !empty($recordSmsOptIn)` here?

app/helpers.php
```diff
@@ -0,0 +1,18 @@
...
+ */
+function str_to_boolean($text)
+{
+    $sanitized = strtolower($text);
+
+    if ($sanitized === 'y') {
+        return true;
+    }
+
+    return filter_var($sanitized, FILTER_VALIDATE_BOOLEAN);
```
chloealee: never knew about `FILTER_VALIDATE_BOOLEAN` - so cool! Definitely want to use and update some things in Rogue with this.

app/Jobs/CreateRockTheVotePostInRogue.php
aaronschachter: I've added the check for `isset`, but looking into [the `empty` PHP docs](http://php.net/manual/en/function.empty.php) -- I'm not sure we'd want to use it per adding the `str_to_boolean` helper in this PR.

Per the docs:
> The following values are considered to be empty:
> - "" (an empty string)
> - 0 (0 as an integer)
> - 0.0 (0 as a float)
> - "0" (0 as a string)
> - NULL
> - FALSE
> - array() (an empty array)

Per our `str_to_boolean`, we're checking for more than just 'no' values.  If there is a 0 set in our column, we'll interpet it as a `false` value to save (vs not saving it at all)

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -211,64 +222,61 @@ private function parseReferralCode($referralCode)
     }
 
     /**
-     * For a given record and referral code values, first check if we have a northstar ID, then grab the user using that.
-     * Otherwise, see if we can find the user with the given email (if it exists), if not check if we can find them with the given phone number (if it exists).
-     * If all else fails, create the user .
+     * Fetch user ID by given email or mobile if exists. If not found, create new user.
      *
-     * @TODO - If we have a northstar id in the referral code, then we probably don't need to make a call to northstar for the full user object.
```
aaronschachter: While revisiting the logic for when to use `isset` and `empty` in checking for variables, I thought it made sense to fix this TODO by moving the check for a `northstar_id` column into `handle` instead of passing a `values` parameter here.

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -39,35 +39,42 @@ public function __construct($record)
      */
     public function handle(Rogue $rogue)
     {
-    	$shouldProcess = $this->scrubRecord($this->record);
+    	$isTestData = $this->isTestData($this->record);
```
DFurnes: This is way clearer. 🎉

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -39,35 +39,42 @@ public function __construct($record)
...
+            $referralCodeValues = $this->parseReferralCode($this->record['Tracking Source']);
+            info('Referral code: ' . implode(', ', $referralCodeValues));
 
-            $user = $this->getOrCreateUser($this->record, $referralCodeValues);
+            $userId = $referralCodeValues['northstar_id'];
+            $campaignId = (int) $referralCodeValues['campaign_id'];
 
-            // @TODO: If we write more functions that are identical across all voter reg imports, pull out into a ImportsVoterReg trait
-            $this->updateNorthstarStatus($user, $this->translateStatus($this->record['Status'], $this->record['Finish with State']));
+            if (!isset($userId) || empty($userId)) {
```
DFurnes: This can just be `if (empty($userId))`, since we know the variable will be defined at this point.

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -39,35 +39,42 @@ public function __construct($record)
...
-            $post = $rogue->getPost([
-                'campaign_id' => (int) $referralCodeValues['campaign_id'],
-                'northstar_id' => $user->id,
-                'type' => 'voter-reg',
+            // @TODO: If we write more functions that are identical across all voter reg imports, pull out into a ImportsVoterReg trait
+            $this->updateNorthstarStatus($userId, $this->translateStatus($this->record['Status'], $this->record['Finish with State']));
+
+            $postType = 'voter-reg';
+            // Check if post exists.
+            $fetchPostRes = $rogue->getPost([
```
DFurnes: Can we rename this variable to `$existingPosts`? I'm not sure it's explicitly stated anywhere, but we tend to avoid abbreviations wherever possible.

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -39,35 +39,42 @@ public function __construct($record)
...
+                'campaign_id' => $campaignId,
+                'northstar_id' => $userId,
+                'type' => $postType,
```
DFurnes: In the future, let's consider updating our Rock The Vote referral string to just pass along an `action_id` (so we don't need to guess using campaign ID & post type). Yay post metadata!

app/Jobs/CreateRockTheVotePostInRogue.php
DFurnes: This is unrelated to this change, but it'd also be nice to rename this method to `$rogue->getPosts`. I didn't realize this was returning an array of posts until I'd read further down this file.

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -211,64 +222,61 @@ private function parseReferralCode($referralCode)
...
+            'email' => $recordEmail,
+            'mobile' => $recordMobile,
+            'first_name' => $record['First name'],
+            'last_name' => $record['Last name'],
+            'addr_street1' => $record['Home address'],
+            'addr_street2' => $record['Home unit'],
+            'addr_city' => $record['Home city'],
+            'addr_state' => $record['Home state'],
+            'addr_zip' => $record['Home zip code'],
+            'source' => env('NORTHSTAR_CLIENT_ID'),
```
DFurnes: Let's read this from `config('services.northstar.client_credentials.client_id')` instead so we don't prevent ourselves from using [config caching](https://laravel.com/docs/5.7/configuration#configuration-caching) here (if we don't already).

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -211,64 +222,61 @@ private function parseReferralCode($referralCode)
...
-                throw new Exception(500, 'Unable to create user: ' . $record['Email address']);
-            }
+        if ($user->id) {
+            info('created user', ['user' => $user->id]);
+        } else {
+            throw new Exception(500, 'Unable to create user: ' . $recordEmail);
         }
 
-        return $user;
+        return $user->id;
```
DFurnes: Is it more helpful here to return the full `NorthstarUser` object in case we wanted to read an existing user's profile fields for something? (I also tend to err on the side of passing & returning objects whenever possible since you can add type-checking then too… ensuring you pass a `User` with a _user_ ID, rather than just any string ID).

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -356,35 +364,18 @@ private function updateStatus($currentStatus, $newStatus)
...
-
     /*
      * Translate to Northstar status and update Northstar user (Northstar takes care of the hierarchy)
      *
-     * @param Object $user
+     * @param string $userId
      * @param string $statusToSend
     */
-    private function updateNorthstarStatus($user, $statusToSend)
+    private function updateNorthstarStatus($userId, $statusToSend)
```
DFurnes: Same thought here: passing a user instance (if we added a `NorthstarUser $user` type hint) would avoid the chance of sneaky bugs if we accidentally passed the wrong kind of ID to this function.

app/Jobs/CreateRockTheVotePostInRogue.php
DFurnes: Yeah, I'd be wary of over-using `isset()` or `empty()` unless you specifically need the functionality they're providing (checking whether a variable is defined, like a nested array reference / checking whether a variable is defined & truthy).

Since they swallow up so many "false" conditions, they could hide sneaky bugs.

aaronschachter: The initial code was simply checking for `if ($record['Phone'])` vs isset, empty -- do we even need the `isset`? I was also looking at `array_key_exists`...but it seems unlikely we'd ever receive a NULL value via importing the csv

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -77,29 +84,33 @@ public function handle(Rogue $rogue)
...
+                    $rogue->updatePost($postId, ['status' => $statusShouldChange]);
                 }
             }
         }
     }
 
     /*
-     * Determines if a record should be process to be stored or if it is not valid.
+     * Returns whether a record is test data that should not create/update users and/or posts.
+     * TODO: Move this into helpers and DRY with any other Jobs that are still relevant.
```
aaronschachter: Outside of the scope of this PR, but should we delete the TurboVote job if it's no longer used? @ngjo

ngjo: Let's hold off for a little while longer -- we get some straggler registrations so while I've used it sparingly, there has been use for it. I created an icebox ticket tho for it https://www.pivotaltracker.com/story/show/164114650

app/Jobs/CreateRockTheVotePostInRogue.php
aaronschachter: I changed this up to avoid fetching a user when the csv already has a `northstar_id` column when completing that TODO in https://github.com/DoSomething/chompy/pull/58/commits/3c09ebded9850571da151d0006a3cdde52eda7aa. It feels odd to wrap a dummy Northstar user object/array around the id.. but also to fetch the user when we already have a NS ID? (Unless we want super sanity checking to ensure the file NS ID is correct)

app/Jobs/CreateRockTheVotePostInRogue.php
DFurnes: Yeah, I don't think we need an `isset` here. 👍

app/Jobs/CreateRockTheVotePostInRogue.php
DFurnes: I think it's a good idea to sanity-check the ID exists in these cases. (We've had issues with bad data being provided to Chompy before, so an invalid Northstar ID doesn't seem out of the question).

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -211,62 +224,72 @@ private function parseReferralCode($referralCode)
...
+     * Creates new user from job record.
+     *
+     * @return NorthstarUser
+     */
+    private function createUser()
+    {
+        $record = $this->record;
+        $userData = [
+            'email' => $this->recordEmail,
+            'mobile' => $this->recordMobile,
```
aaronschachter: Does this approach work? Looking to avoid hardcoding the `record['Email address']` and `record['Phone']` array keys everywhere with the `$this->recordEmail` and `$this->recordMobile` properties... but then also doesn't seem right to even pass the `$record` parameter into this function at all (as we did before) - which is why I've removed.

DFurnes: I wonder if it makes sense to parse the whole CSV record into a sanitized in-memory representation first, so instead of referencing `$this->recordEmail` or `$this->record['Email address']` everywhere, we handle all that messy business up front:

```php
class RockTheVoteRecord {
    public function __construct($record)
    {
        $this->email = $record['Email address'];
        $this->mobile = $record['Phone'];
        $this->first_name = $record['First name'],
        // and so on...
    }
}

class CreateRockTheVotePostInRogue implements ShouldQueue
    public function __construct($record)
    {
        $this->record = new RockTheVoteRecord($record);
    }

    // then anywhere we need these properties, we can use
    // the "standard" names we parsed above…
```

aaronschachter: I like this idea... but also want to avoid adding even more changes outside of the initial scope of this PR... as that gets even closer to rewriting the entire file. Maybe I'll add a `RockTheVote` class per your example with `record, `email` and `mobile` properties, with a TODO to parse the rest in a future PR?

DFurnes: Yeah, that makes good sense to me!

PR #59, opened 2019-02-13
Persist selected import type on webform

aaronschachter:
#### What's this PR do?

Branching off of #58, this PR refactors the import form to post the `import-type` parameter via a query string, set by new nav links for each import type. As I was testing #58, I thought it'd be nice to get to know the app a bit better (and help make it  slightly less tedious to repeatedly test) by fixing the TODO for pre-selecting an import type.

screen shot 2019-02-13 at 3 05 50 pm

* Verify that logging in directs you to the homepage, and that the "Import" link/form is separated into three links per import type. 

* Verify import form works per each selected import type, and that you are redirected to the same import type form after successful submit

* Verify import form is hidden if type query parameter is not one of `turbovote`, `rock-the-vote`, or `facebook`

#### Relevant tickets

Support to make testing https://www.pivotaltracker.com/story/show/163564896 easier

### DIFF
--- a/app/Http/Controllers/ImportController.php
+++ b/app/Http/Controllers/ImportController.php
@@ -58,17 +58,17 @@ class ImportController extends Controller
 
         $type = $request->input('import-type');
 
-        if ($type === 'turbovote') {
+        if ($type === get_import_type_turbovote()) {
             info("turbo vote import happening");
             ImportTurboVotePosts::dispatch($path)->delay(now()->addSeconds(3));
         }
 
-        if ($type === 'rock-the-vote') {
+        if ($type === get_import_type_rock_the_vote()) {
             info('rock the vote import happening');
             ImportRockTheVotePosts::dispatch($path)->delay(now()->addSeconds(3));
         }
 
-        if ($type === 'facebook') {
+        if ($type === get_import_type_facebook()) {
             info("Facebook share import happening");
             ImportFacebookSharePosts::dispatch($path)->delay(now()->addSeconds(3));
         }
--- a/app/Http/Controllers/ImportController.php
+++ b/app/Http/Controllers/ImportController.php
@@ -58,17 +58,17 @@ class ImportController extends Controller
 
         $type = $request->input('import-type');
 
-        if ($type === get_import_type_turbovote()) {
+        if ($type === \Chompy\ImportType::$turbovote) {
             info("turbo vote import happening");
             ImportTurboVotePosts::dispatch($path)->delay(now()->addSeconds(3));
         }
 
-        if ($type === get_import_type_rock_the_vote()) {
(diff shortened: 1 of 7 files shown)


app/helpers.php
```diff
@@ -1,7 +1,47 @@
 <?php
 
 /**
- * Parse a CSV field value as boolean.
+ * Returns import type for Facebook share files.
+ * @return string
+ */
+function get_import_type_facebook()
```
DFurnes: Let's make a `ImportType` class that exports these constants:

```php
// app/Import.php
namespace Chompy;

class ImportType {
  public static $facebook = 'facebook';
  public static $rockTheVote = 'rock-the-vote';
  public static $turbovote = 'turbovote';

  public static function all() {
    return [self::$facebook, self::$rockTheVote, self::$turbovote];
  }
}
```

We can then use these anywhere else with `\Chompy\ImportType::$turbovote` or `\Chompy\ImportType::all()` without cluttering our helpers file with tons of constants over time.

resources/views/components/nav.blade.php
```diff
@@ -13,8 +13,24 @@
         <div id="navbar" class="navbar-collapse collapse">
             <ul class="nav navbar-nav navbar-right">
                 @if (Auth::user())
-                    <li><a class="nav-item nav-link" href="/import">Import</a></li>
-                    <li><a class="nav-item nav-link" href="/logout">Logout</a></li>
+                    <li class="{{ (app('request')->input('type') === get_import_type_turbovote()) ? 'active' : '' }}">
```
DFurnes: Totally subjective, but you can use the `request()` helper here as a shortcut to `app('request')`.

app/Http/Controllers/ImportController.php
```diff
@@ -56,21 +56,23 @@ public function store(Request $request)
             throw new HttpException(500, 'Unable read and store file to S3.');
         }
 
-        if ($request->input('import-type') === 'turbovote') {
+        $type = $request->input('import-type');
+
+        if ($type === \Chompy\ImportType::$turbovote) {
```
DFurnes: If you want, you can import this class at the top of the file:

```php
use Chompy\ImportType;
```

and then these can be shortened a wee bit, e.g:

```php
if ($type === ImportType::$turbovote) {
```

app/Import.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Chompy;
+
+class ImportType {
+  public static $facebook = 'facebook';
```
DFurnes: Would you mind adding docblocks for each of these properties and the method before merging? e.g.

```php
/**
 * A Facebook share import type.
 *
 * @var string
 */
public static $facebook = 'facebook';

/**
 * Return the list of valid import types.
 *
 * @return array
 */
public static function all() {
```

PR #60, opened 2019-02-23
RTV Cleanup: Part 1

aaronschachter:
#### What's this PR do?

Moves parsing a referral code into the  `RockTheVoteRecord` constructor introduced in #58, to improve maintainability by separating parsing CSV's values out from the Northstar and Rogue API requests executed in the `CreateRockTheVotePostInRogue` handler.

Does a bit of renaming to improve readability (it took me a long time to figure out what's going on in each):
* Renames `translateStatus` to `parseVoterRegistrationStatus`
* Renames `updateStatus` to `getVoterRegistrationStatusChange`

Also moves the logic for identifying test email addresses into an `is_valid_email` helper.

Verify import works as expected, no changes should have been made. 

This adds a TODO cleanup -- after creating a new user, we make a 2nd request to update their `voter_registration_status`. This should be refactored to avoid making the second update request.

### DIFF
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -15,16 +15,21 @@ class RockTheVoteRecord {
     public function __construct($record)
     {
         $this->post_type = 'voter-reg';
+        $this->source = 'rock-the-vote';
+
         $this->email = $record['Email address'];
         $this->mobile = $record['Phone'];
         $this->email_opt_in = $record['Opt-in to Partner email?'];
         // Note: Not a typo, this column name does not have the trailing question mark.
         $this->sms_opt_in = $record['Opt-in to Partner SMS/robocall'];
 
+        $this->voter_registration_status = $this->parseVoterRegistrationStatus($record['Status'], $record['Finish with State']);
+
         $referral = $this->parseReferralCode($record['Tracking Source']);
         $this->user_id = $referral['northstar_id'];
         $this->campaign_id = (int) $referral['campaign_id'];
         $this->source_details = $referral['source_details'];
+
         // TODO: Add all other properties used for getting/validating/creating users/posts.
     }
 
@@ -107,7 +112,7 @@ class RockTheVoteRecord {
                     'northstar_id' => null, // set the user to null so we force account creation when the code is not present.
                     'campaign_id' => $values['campaign_id'],
                     'campaign_run_id' => $values['campaign_run_id'],
-                    'source' => 'rock-the-vote',
+                    'source' => $this->source,
                     'source_details' => null,
                 ];
             }
@@ -118,7 +123,7 @@ class RockTheVoteRecord {
                     'northstar_id' => $values['northstar_id'], // set the user to null so we force account creation when the code is not present.
                     'campaign_id' => 8017,
                     'campaign_run_id' => 8022,
-                    'source' => 'rock-the-vote',
+                    'source' => $this->source,
                     'source_details' => null,
                 ];
             }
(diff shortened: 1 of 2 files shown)


app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -227,41 +135,155 @@ private function parseReferralCode($referralCode)
...
-     * If userId param is given, check for valid user.
-     * Otherwise check by job record email, mobile.
+     * Execute the job to create a Turbo Vote post in Rogue.
+     *
+     * @return array
+     */
+    public function handle(Rogue $rogue)
+    {
+        if (is_test_email($this->data->email)) {
+            info('progress_log: Skipping test: ' . $this->data->email);
```
aaronschachter: @ngjo - This PR keeps the logic for checking for test emails, but removes the logic of looking for last names that include `Baloney`. Do we need to keep the check for test last name?

ngjo: @aaronschachter yea, I think we can remove that check for `Baloney` now

DFurnes: Oh jeeeeez haha, yeah glad we're removing that. 😂

app/helpers.php
```diff
@@ -1,5 +1,18 @@
...
+
+/**
+ * Returns whether email is a test email.
+ *
+ * @param string $email
+ * @return boolean
+ */
+function is_test_email($email)
+{
+    return Str::contains($email, ['thing.org', '@dosome','rockthevote.com', 'test', '+']);
```
DFurnes: I wonder what the thinking was behind these separate `thing.org` and `@dosome` checks… can we explicitly check for full `@dosomething.org` and `@example` emails instead? (I believe that's how we filter out test users in Quasar.)

I'm also a little concerned about the overly generous `test` and `+` checks. For example, thinking how this might trigger unexpectedly on a user like "Mitt Ester" with the email `mittester@gmail.com`, or folks who like to use Google's [special `+` logic](https://gizmodo.com/how-to-use-the-infinite-number-of-email-addresses-gmail-1609458192) to filter out marketing emails.

(I know this is just refactoring the existing logic into a helper, but it caught my eye!)

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -227,41 +135,155 @@ private function parseReferralCode($referralCode)
...
+
+        if (strpos($rtvStatus, 'step') !== false) {
+            return 'uncertain';
+        }
+
+        if ($rtvStatus === 'rejected' || $rtvStatus === 'under 18') {
+            return 'ineligible';
+        }
+
+        return '';
```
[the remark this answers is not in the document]
aaronschachter: I really can't say, as I don't know the history of the code.. but it seems safe to simply not update the user's registration status, vs throw the error? 🤷‍♂️

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -204,7 +112,7 @@ private function parseReferralCode($referralCode)
                     'northstar_id' => null, // set the user to null so we force account creation when the code is not present.
                     'campaign_id' => $values['campaign_id'],
```
aaronschachter: I am 🤔why a post is created for 8017 instead of 8167 in this test file Tracking Code`user:null,campaign:8167,campaignRunId:8168,source:web,source_details:PollCallActionPage `.

I'm fairly certain #61 is throwing an error for me locally because it's trying to create a post for campaign 8167. I'm also wondering if some of this logic has been deprecated and never removed?

chloealee: @aaronschachter hm this is weird! Are you testing on QA? We don't do anything with `campaign_run_id` in Rogue anymore, just look for `campaign_id`...

PR #61, opened 2019-02-23
RTV Cleanup: Part 2

aaronschachter:
#### What's this PR do?

Branches from #60, which was opened first to help keep these reviewable, to finish cleanup, moving all record parsing into the `RockTheVoteRecord` constructor.

* Refactors `parseReferralCode` as `parseUserId` - as that's the only value we're using for import per https://github.com/DoSomething/chompy/pull/61#discussion_r260439032
 
* Renames `extractDetails` as `parsePostDetails`

* Refactors `updateNorthstarStatus` as `updateUser`

* Removes `campaign_run_id` references

Verify expected behavior 👀

### DIFF
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -35,7 +35,7 @@ class RockTheVoteRecord {
             $this->sms_status = str_to_boolean($smsOptIn) ? 'active' : 'stop';
         }
 
-        $rtvStatus = $this->parseRtvStatus($record['Status'], $record['Finish with State']);
+        $rtvStatus = $this->parseVoterRegistrationStatus($record['Status'], $record['Finish with State']);
        
         $this->voter_registration_status =  Str::contains($rtvStatus, 'register') ? 'registration_complete' : $rtvStatus;
 
@@ -103,7 +103,7 @@ class RockTheVoteRecord {
      * @param  string $rtvFinishWithState
      * @return string
      */
-    private function parseRtvStatus($rtvStatus, $rtvFinishWithState)
+    private function parseVoterRegistrationStatus($rtvStatus, $rtvFinishWithState)
     {
         $rtvStatus = strtolower($rtvStatus);
 
@@ -222,7 +222,7 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
     }
 
     /**
-     * Check for user first by id, next by email, last by mobile.
+     * Check for user first by record user_id, next by email, last by mobile.
      *
      * @param array $record
      * @return NorthstarUser
@@ -248,14 +248,17 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
     }
 
     /**
-     * Creates new user.
+     * Creates new user from record.
      *
+     * @param array $record
      * @return NorthstarUser
      */
     private function createUser($record)
     {
         $userData = [];
+
         $userFields = ['addr_city', 'addr_state', 'addr_street1', 'addr_street2', 'addr_zip', 'email', 'mobile', 'first_name', 'last_name', 'voter_registration_status'];
+
         foreach ($userFields as $key) {
             $userData[$key] = $record->{$key};
         }
@@ -265,6 +268,7 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
         if (!empty($record->email_subscription_status)) {
             $userData['email_subscription_status'] = $record->email_subscription_status;
         }
+
         if (!empty($record->sms_status)) {
             $userData['sms_status'] = $record->sms_status;
         }
@@ -279,7 +283,7 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
         return $user;
     }
 
-    /*
+    /**
      * Determines if a status should be changed and what it should be changed to.
      *
      * @param string $currentStatus
@@ -302,8 +306,8 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
         return $indexOfCurrentStatus < $indexOfNewStatus ? $newStatus : null;
     }
 
-    /*
-     * Translate to Northstar status and update Northstar user (Northstar takes care of the hierarchy)
+    /**
+     * Update Northstar user with given data.
      *
      * @param Object $user
      * @param Array $data
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -24,6 +24,7 @@ class RockTheVoteRecord {
         $this->first_name = $record['First name'];
         $this->last_name = $record['Last name'];
         $this->mobile = $record['Phone'];
+        $this->source = config('services.northstar.client_credentials.client_id');
 
         $emailOptIn = $record['Opt-in to Partner email?'];
         if ($emailOptIn) {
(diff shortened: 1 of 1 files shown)


app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -52,98 +70,34 @@ private function parseReferralCode($referralCode)
...
 
-                    default:
-                        break;
-                }
+        $referralCode = explode(',', $referralCode);
 
-                // Is this a referral?
-                if (strtolower($value[0]) === 'referral' && strtolower($value[1]) === 'true') {
-                    $referral = true;
-                }
```
aaronschachter: Reading up on the docs for [Referral Links](https://github.com/DoSomething/chompy/wiki/How-Rock-The-Vote-(RTV)-records-are-imported#special-case-referral-links), it seems we would not would want to remove checking for a `referral` query parameter. However we'll see later when looking for `$referral`...

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -52,98 +70,34 @@ private function parseReferralCode($referralCode)
...
+            if ($key === 'user') {
+                $values['user_id'] = $value[1];
+            }
+            if (($key === 'campaignid' || $key === 'campaign') && is_numeric($value[1])) {
+                $values['campaign_id'] = (int) $value[1];
             }
         }
 
-        // If we were missing all the necessary values or if this is a referral, use all the defaults
-        if (empty($finalValues) || $referral) {
```
aaronschachter: We only reference  the `$referral` variable removed in https://github.com/DoSomething/chompy/pull/61/files#r259975670... `$referral` here, when we're simply returning the parsed referral campaign as 8017.

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -52,98 +70,34 @@ private function parseReferralCode($referralCode)
...
-                    'campaign_run_id' => 8022,
-                    'source' => $this->source,
-                    'source_details' => null,
-                ];
+            $key = strtolower($value[0]);
+            if ($key === 'user') {
+                $values['user_id'] = $value[1];
+            }
+            if (($key === 'campaignid' || $key === 'campaign') && is_numeric($value[1])) {
+                $values['campaign_id'] = (int) $value[1];
```
aaronschachter: Is there a need currently to send campaign ID from a referral string? Shouldn't this just be a server-side config variable per https://dosomething.slack.com/archives/CA0N5FXND/p1551114468011700?thread_ts=1551114281.011600&cid=CA0N5FXND

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -10,27 +10,45 @@
...
 
         $referral = $this->parseReferralCode($record['Tracking Source']);
-        $this->user_id = $referral['northstar_id'];
-        $this->campaign_id = (int) $referral['campaign_id'];
-        $this->source_details = $referral['source_details'];
-
-        // TODO: Add all other properties used for getting/validating/creating users/posts.
+        $this->user_id = !empty($referral['user_id']) ? $referral['user_id'] : null;
+        // TODO: Do we need to check the referral for these values? Don't seem they are ever used.
+        $this->campaign_id = 8017;
```
chloealee: is this hardcoding the campaign so it is `8017` each time, since it happens after `parseReferralCode`?

aaronschachter: It is -- it's essentially ignoring the campaign found in `parseReferralCode`. I added this in, since I wasn't sure if the desired behavior was to create a post for 8017, or for the campaign found in a referral code... which is related to thread in https://github.com/DoSomething/chompy/pull/60#discussion_r259963459

chloealee: I might have been misunderstanding your Q the whole time! Are you asking for the context for why we want to default to `8017` each time instead of getting what is in `parseReferralCode`?

aaronschachter: I was! @ngjo and I chatted through the history of these imports today, and landed on the requirements we need for present day referral codes:

* When fetching a user,  use the parsed `user` referral parameter only when the parsed `referral` referral parameter exists and is set to `true` (otherwise, check for user by email, then mobile)

* Ignore the `campaign_id`, `campaign_run_id`, `source`, and `source_details` referral parameters completely.

    * When creating the post, send an action ID which will be set via hardcoding or config var -- once https://github.com/DoSomething/rogue/pull/837 is deployed, refs https://github.com/DoSomething/chompy/pull/58/files#r258541121 and [Slack thread](https://dosomething.slack.com/archives/CA0N5FXND/p1551114281011600)

    * Continue to send hardcoded post source as `rock-the-vote`, `source_details` as null

* We will continue to send the post `details` as is (renamed as `parsePostDetails` in this PR) 

I'll update this PR description and ping folks when this ready for 👀 . Thanks for taking a look @chloealee !

chloealee: This is a great summary and so much clearer than what we have as-is! Thanks for taking the time to get to the bottom of this and thanks for @ngjo for clearing up! My memory is so bad! 👵 Will keep an eye out for your notification when it is ready for another look!

aaronschachter:
This PR has been updated per https://github.com/DoSomething/chompy/pull/61#discussion_r260439032 and ready for review whenever anyone has a chance.

app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -10,141 +10,108 @@
...
+                /**
+                 * Return null to force querying for existing user via this record email or mobile
+                 * upon import.
+                 */
+                return null;
+            }
         }
 
-        return $finalValues;
+        return isset($userId) ? $userId : null;
```
DFurnes: The `isset()` here gives me a little pause (since as a reader, I have to then Cmd+F to see where it may or may not get set above). Can we instead order the conditionals so this reads "in order"? e.g.

```php
    // We expect 'referral', but check for any typos in any manually entered URLs.
    if (($key === 'referral' || $key === 'refferal') && str_to_boolean($value[1])) {
        return null;
    }

    // We expect 'user', but check for any variations/typos in any manually entered URLs.
    if ($key === 'user' || $key === 'user_id' || $key === 'userid') {
        return $value[1];
    }
}

return null;
```

aaronschachter: I didn't return the `userId` if found -- to handle an edge case where an editor may add the referral parameter after the user parameter, e.g. `puppetsloth.org?user=5547be89469c64ec7d8b518d&referral=true` -- In this case we want to return `null` for a user ID

aaronschachter: We could set `$userId` to `null` in the beginning of the function? I had that at first, replacing `$values` -- but then went this route

DFurnes: Oh! I didn't notice this was inside a `foreach`! 🤦‍♂️ 

It might make sense to make a first pass to "read" the referral string into an array (e.g. `user=5547be89469c64ec7d8b518d&referral=true` into `['user' => '…', 'referral' => true]`, since then you can have more "straightforward" conditions without having to worry about ordering. But definitely not a big thing!

aaronschachter: Sounds good - we can revisit when ready to send over an `action_id` instead of `campaign_id`

PR #64, opened 2019-03-08
Callpower endpoint

chloealee:
#### What's this PR do?
This PR adds an endpoint to integrate CallPower directly with our system. 
- The endpoint is hidden behind a third party static API Key that CallPower will hit ~every hour to send new records in the database. 
- We receive general information from CallPower - this endpoint handles most of the logic to prepare the data in a way we'd like it to be received in Rogue to create the post and/or signup. 
- Adds `.editorconfig` file.

- Does all logic make sense? 
- Anything I'm missing that is outlined in the [tech spec](https://docs.google.com/document/d/1aDNLKZ6BPto65Ce5RDJkXf8LpKGIFdftO13WgZgkO48/edit)? 

#### Relevant tickets

Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/164476751) Pivotal Card.


routes/api.php
```diff
@@ -16,3 +16,14 @@
 Route::middleware('auth:api')->get('/user', function (Request $request) {
     return $request->user();
 });
+
+// Third Party Integration
+$router->group(['prefix' => 'api/v1', 'middleware' => ['third-party-api-token-auth']], function() {
```
DFurnes: The `routes/api.php` file is [already prefixed](https://github.com/DoSomething/chompy/blob/85b19765da65305192fea67f199642c1b980b4b2/app/Providers/RouteServiceProvider.php#L66-L72) with `api/`, so this should just read `'prefix' => 'v1'`.

chloealee: omg THANK YOU!!

app/Http/Controllers/ThirdParty/CallPowerController.php
```diff
@@ -0,0 +1,139 @@
...
+{
+
+    /**
+     * Create a controller instance.
+     *
+     * @return void
+     */
+    public function __construct(Rogue $rogue)
+    {
+      // @TODO: add role middleware?
```
chloealee: Do we need to add role middleware here or will the partner specific static API key be sufficient here?

app/Http/Controllers/ThirdParty/CallPowerController.php
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     *
+     * @param Request $request
+     */
+    public function store(Request $request)
+    {
+    	$request->validate([
+    		// QUESTION - We shouldn't have to validate phone number because this will come from Twilio and phone numbers are validated in NS?
+            'mobile' => 'required',
```
chloealee: We shouldn't have to validate phone number because this will come from Twilio and phone numbers are validated in NS, right?

app/Http/Controllers/ThirdParty/CallPowerController.php
```diff
@@ -0,0 +1,139 @@
...
+            'callpower_campaign_id' => 'required|integer',
+            // QUESTION: do we want to make it this rigid? What if CallPower statuses change?
+            'status' => 'required|string|in:completed,busy,failed,no answer,cancelled, unknown',
```
chloealee: Do we want to make sure the `status` is within `completed`, `busy`, `failed`, `no answer`, cancelled`, or `unknown` (possible CallPower statuses)? I'm leaning towards no because what if CallPower statuses change? Also, if it is anything other than `completed`, we'll just mark the post as `incomplete`, with status details in the `details` of a post. If it is `completed`, the post's `status` will be `accepted`.

app/Http/Controllers/ThirdParty/CallPowerController.php
```diff
@@ -0,0 +1,139 @@
...
+    	$keys = [
+    		'status_details',
+    		'call_timestamp',
+    		'call_duration',
+    		'campaign_target_name',
+    		'campaign_target_title',
+    		'campaign_target_district',
+    		'callpower_campaign_name',
+    		'number_dialed_into',
+    	];
```
chloealee: @sohaibhasan @mjain-1 FYI here are the working names for how data outlined in the [CallPower / Rogue Integration Technical Spec](https://docs.google.com/document/d/1aDNLKZ6BPto65Ce5RDJkXf8LpKGIFdftO13WgZgkO48/edit#heading=h.p6clhgy4rvbk) will be saved in a post's `details`

sohaibhasan: Love it!

app/Http/Controllers/ThirdParty/CallPowerController.php
DFurnes: Yeah, I think that's a good assumption!

app/Http/Controllers/ThirdParty/CallPowerController.php
DFurnes: Yeah, agreed that using less validation here is probably better (since we can then always re-try the particular record in Chompy if it fails from downstream validation in Northstar or Rogue).

app/Http/Controllers/ThirdParty/CallPowerController.php
DFurnes: Yeah, we'd only use `role` middleware if we're getting an OAuth token (with a user or machine role attached). There's no real concept of a "role" when taking in a static key like this.

app/Http/Controllers/ThirdParty/CallPowerController.php
```diff
@@ -0,0 +1,134 @@
...
+            'call_duration' => 'required|integer',
+            'campaign_target_name' => 'required|string',
+            'campaign_target_title' => 'required|string',
+            'campaign_target_district' => 'required|string',
+            'callpower_campaign_name' => 'required|string',
+            'number_dialed_into' => 'required',
+        ]);
+
+        // Using the mobile number, get or create a northstar_id.
+        $user = $this->getOrCreateUser($request['mobile']);
```
DFurnes: This should all be happening in a queued job, like we do for other imports.

app/Http/Middleware/AuthenticateWithApiKey.php
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     *
+     * @throws \Illuminate\Auth\AuthenticationException
+     */
+    public function handle($request, Closure $next)
+    {
+        // Only protected POST, PUT, PATCH, and DELETE routes
```
DFurnes: This conditional (to "only protected POST, PUT, PATCH, and DELETE routes") can be removed.

app/Http/Controllers/ThirdParty/CallPowerController.php
```diff
@@ -0,0 +1,134 @@
...
+       	$existingPost = $this->rogue->getPost([
+            'action_id' => $action['data'][0]['id'],
+            'northstar_id' => $user->id,
+        ]);
+        if (! $existingPost['data']) {
+            info('creating post in rogue for northstar user: ' . $user->id);
+
+            $details = $this->extractDetails($request);
+
+        	// Determine source details.
```
DFurnes: Something looks wrong with indentation here.

app/Http/Controllers/ThirdParty/CallPowerController.php
```diff
@@ -0,0 +1,134 @@
...
+        		'northstar_id' => $user->id,
+        		'action_id' => $action['data'][0]['id'],
+        		'type' => 'phone-call',
+        		'status' => $request['status'] === 'completed' ? 'accepted' : 'incomplete',
+        		'quantity' => 1,
+        		'source' => 'CallPower',
+        		'details' => $details,
+        	]);
+
+        	if ($post['data']) {
```
DFurnes: Same, indentation looks wonky. Can you re-format this file with your editor?

app/Http/Controllers/ThirdParty/CallPowerController.php
```diff
@@ -0,0 +1,134 @@
...
+    {
+    	// Get the user by mobile number.
+    	info('getting user with the mobile: ' . $mobile);
+        $user = gateway('northstar')->asClient()->getUser('mobile', $mobile);
+
+    	// If there is no user, create one.
+    	if (is_null($user)) {
+    		$user = gateway('northstar')->asClient()->createUser([
+                'mobile' => $mobile,
+                'source' => env('NORTHSTAR_CLIENT_ID'),
```
DFurnes: This should be reading from `config('services.northstar.client_credentials.client_id')` if we need to access this environment variable (see [note about config caching](https://laravel.com/docs/5.8/deployment#optimizing-configuration-loading) from the docs). I think this can (and should) also be excluded from this API call, since Northstar will set it automatically if not provided in the request.

app/Http/Controllers/ThirdParty/CallPowerController.php
```diff
@@ -0,0 +1,134 @@
...
+            ]);
+    	}
+
+    	// Log if the user was successfully created.
+            if ($user->id) {
```
DFurnes: More weird indentation here. 👾

app/Http/Controllers/ThirdParty/CallPowerController.php
```diff
@@ -0,0 +1,134 @@
...
+            'call_timestamp',
+            'call_duration',
+            'campaign_target_name',
+            'campaign_target_title',
+            'campaign_target_district',
+            'callpower_campaign_name',
+            'number_dialed_into',
+        ];
+
+        foreach ($keys as $key) {
```
DFurnes: Instead of building this object using a loop, can we just inline to make it easier to read?

```php
return json_encode([
  'status_details' => $call['status'],
  'call_timestamp' => $call['call_timestamp'],
  'call_duration' => $call['call_duration'],
  // ...
]);
```

app/Http/Controllers/ThirdParty/CallPowerController.php
chloealee: @DFurnes I don't think that Chompy does this? For RTV, seems like it does the same thing as it does [here](https://app.ghostinspector.com/tests/5c49f1dd638e692a231ece5f)?

app/Http/Controllers/ThirdParty/CallPowerController.php
chloealee: oh yes this is so much better! Definitely over thought that.

app/Http/Controllers/ThirdParty/CallPowerController.php
DFurnes: I think you might have pasted the wrong link in there?

This is how our other imports in Chompy work, for example, here's where [we kick off queued jobs for RTV imports in `ImportsController`](https://github.com/DoSomething/chompy/blob/a08a67897c1835846cae30667b9e1d6f7e5eff26/app/Http/Controllers/ImportController.php#L62-L75). We do this because otherwise the whole "import" has to happen before Chompy can render a response, and we'd run into request timeouts for anything more than the shortest of CSVs.

This endpoint here, admittedly, is not parsing a whole CSV! But we still don't want _all_ this work (let's say, 200-1000ms to get or create the Northstar user, and 500-1000ms to create the post) to take up time on the main process. Chompy has a limited number of requests it can handle at once, and so if we have really "hefty" requests like this, we'd potentially use up all our resources on something that we don't need a response from right away.

By putting this work into a queued job, we also automatically gain retries for "free" (so if, for example, we find out that CallPower was sending data that caused Northstar or Rogue to error out, we could fix the bug in our systems and simply retry the failing job, rather than having to ask Josh to re-trigger the event on their end).

chloealee: oh I see what you're saying now, thank you for this amazing description @DFurnes ! 

So this should ALL live in a job, not just the `getOrCreateUser` function?

DFurnes: > So this should ALL live in a job, not just the `getOrCreateUser` function?

Yes! You can use [`ImportController`](https://github.com/DoSomething/chompy/blob/a08a67897c1835846cae30667b9e1d6f7e5eff26/app/Http/Controllers/ImportController.php) and the [`CreateRockTheVotePostsInRogue`](https://github.com/DoSomething/chompy/blob/a08a67897c1835846cae30667b9e1d6f7e5eff26/app/Jobs/CreateRockTheVotePostInRogue.php) files as an example for how this should be broken up.

chloealee: awesome, this makes so much sense, thank you again, @DFurnes ! This is updated!

app/Jobs/CreateCallPowerPostInRogue.php
```diff
@@ -0,0 +1,117 @@
...
+        	// Determine source details.
+    		$post = $rogue->createPost([
+    			'northstar_id' => $user->id,
+    			'action_id' => $action['data'][0]['id'],
+    			'type' => 'phone-call',
+    			'status' => $request['status'] === 'completed' ? 'accepted' : 'incomplete',
+    			'quantity' => 1,
+    			'source' => 'CallPower',
+    			'details' => $details,
+    			'dont_send_to_blink' = true,
```
chloealee: We definitely want this for historical records but not sure what we want to do for go-forward. Going to touch base with Jen when she is back from vacation!

app/Jobs/CreateCallPowerPostInRogue.php
```diff
@@ -0,0 +1,117 @@
...
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+
+
+class CreateCallPowerPostInRogue implements ShouldQueue
+{
+	use Dispatchable, InteractsWithQueue, Queueable;
+
+	/**
```
DFurnes: Indentation is off here! ⬅️

app/Jobs/CreateCallPowerPostInRogue.php
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+    /**
+     * Execute the job to create a CallPower post in Rogue.
+     *
+     * @return array
+     */
+    public function handle(Rogue $rogue, Request $request)
+    {
+		// Using the mobile number, get or create a northstar_id.
```
DFurnes: And here! ⬅️

app/Http/Controllers/ThirdParty/CallPowerController.php
```diff
@@ -0,0 +1,43 @@
...
+            'call_duration' => 'required|integer',
+            'campaign_target_name' => 'required|string',
+            'campaign_target_title' => 'required|string',
+            'campaign_target_district' => 'required|string',
+            'callpower_campaign_name' => 'required|string',
+            'number_dialed_into' => 'required',
+        ]);
+
+        // Send to queued job.
+        CreateCallPowerPostInRogue::dispatch($request);
```
DFurnes: Serializing the entire HTTP request in to the queue job feels a little odd to me, since the actual "request" will be done by the time the job picks it up. Instead, I wonder if we should grab the validated array & pass that along?

```php
$parameters = $request->validate([
    // ...
]);

CreateCallPowerPostInRogue::dispatch($parameters);
```

app/Jobs/CreateCallPowerPostInRogue.php
```diff
@@ -0,0 +1,117 @@
...
+    	$user = $this->getOrCreateUser($request['mobile']);
+
+        // Using the callpower_campaign_id, get the action_id from Rogue.
+    	$action = $rogue->getActionFromCallPowerCampaignId($request['callpower_campaign_id']);
+
+        // Check if the post exists in Rogue. If not, create the post.
+    	$existingPost = $rogue->getPost([
+    		'action_id' => $action['data'][0]['id'],
```
DFurnes: It seems a little strange to have to grab this nested `['data'][0]` path from a variable named "action". I think it would be a little cleaner to parse out the action itself in `getActionFromCallPowerCampaignId`, so we don't have to remember the JSON format each time we use this.

app/Jobs/CreateCallPowerPostInRogue.php
```diff
@@ -0,0 +1,117 @@
...
+    		$details = $this->extractDetails($request);
+
+        	// Determine source details.
+    		$post = $rogue->createPost([
+    			'northstar_id' => $user->id,
+    			'action_id' => $action['data'][0]['id'],
+    			'type' => 'phone-call',
+    			'status' => $request['status'] === 'completed' ? 'accepted' : 'incomplete',
+    			'quantity' => 1,
+    			'source' => 'CallPower',
```
DFurnes: I think this should be in the `source_details` column (so that `source` can tell us the application source, Chompy, per our convention).

app/Services/Rogue.php
```diff
@@ -99,4 +98,23 @@ public function updatePost($postId, $input)
...
+     * @return array $action
+     */
+    public function getActionFromCallPowerCampaignId($callpowerCampaignId)
+    {
+        $action = $this->asClient()->get('v3/actions', [
+            'filter' => ['callpower_campaign_id' => $callpowerCampaignId],
+        ]);
+
+        if (! $action['data']) {
+            throw ValidationException::withMessages('Unable to get action data for CallPower campaign id : ' . $callpowerCampaignId);
```
DFurnes: Since this is happening outside the context of a HTTP request, should this just be a plain old `Exception`?

app/Jobs/CreateCallPowerPostInRogue.php
chloealee: ah sorry about all of these @DFurnes ! This is so weird though - I reindented in my editor and it doesn't look like this there! Or am I missing something and need new 👀 ?

DFurnes:
@chloealee I enabled StyleCI on this repository, which will hopefully help with catching some of these indentation problems. It'd also be a good idea to copy over [Rogue's `.editorconfig`](https://github.com/DoSomething/rogue/blob/master/.editorconfig) to this app to automatically configure your indentation settings.

app/Jobs/CreateCallPowerPostInRogue.php
DFurnes: This is probably due to indentation settings being mismatched in your editor – see my comment below about copying over `.editorconfig` & double-check you have [the Sublime Text plugin](https://github.com/sindresorhus/editorconfig-sublime).

aaronschachter:
Nothing needed for merge, but maybe for go live: I'm curious about how we'll be able to monitor how often requests are made to this endpoint. I suppose we could check Papertrail logs, but I could see adding a DB table to log requests made and expose this in the web, so it'd be easy to check - oh wow we haven't gotten any Call Power requests for 5 days, or wow we got 100 yesterday

app/Jobs/CreateCallPowerPostInRogue.php
chloealee: Thanks @DFurnes ! I thought I had that but looks like I don't when I go to `Preferences>Browse Packages...`:
Screen Shot 2019-03-22 at 12 28 42 PM
When I try to download it, only this package shows up:
Screen Shot 2019-03-22 at 12 28 15 PM

DFurnes: I'm not a Sublime Text user so not really sure. 😕 Maybe @weerd might know!

app/Jobs/CreateCallPowerPostInRogue.php
chloealee: oh great call, updated!

app/Http/Controllers/ThirdParty/CallPowerController.php
chloealee: ah yeah I thought this was a little weird too - this is a lot better! Updated :)

app/Jobs/CreateCallPowerPostInRogue.php
chloealee: ah yes way better!

chloealee:
@aaronschachter I love that idea! Do you think if we set up a StatHat alert to push to Slack (similar to what we have for Rogue) that would be a good solution?

chloealee:
@DFurnes ok a lot more files are updated bc of StyleCi updates! One thing though - it says we don't need to include `use Chompy\Jobs\CreateRockTheVotePostInRogue;` in `app/Jobs/ImportRockTheVotePosts.php` and `use Chompy\Jobs\CreateTurboVotePostInRogue;` in `app/Jobs/ImportTurboVotePosts.php` but I think we do?

DFurnes:
> @DFurnes ok a lot more files are updated bc of StyleCi updates! One thing though - it says we don't need to include `use Chompy\Jobs\CreateRockTheVotePostInRogue;` in `app/Jobs/ImportRockTheVotePosts.php` and `use Chompy\Jobs\CreateTurboVotePostInRogue;` in `app/Jobs/ImportTurboVotePosts.php` but I think we do?

That's correct! Since both those classes are in the `\Chompy\Jobs` namespace, you don't need to import them – only if you were trying to use a class from another namespace.

chloealee:
oh got it, thanks @DFurnes ! Ok finally ready for hopefully the last review!

app/Services/Rogue.php
```diff
@@ -99,4 +97,23 @@ public function updatePost($postId, $input)
...
+     * @param int $callpowerCampaignId
+     * @return array $action
+     */
+    public function getActionIdFromCallPowerCampaignId($callpowerCampaignId)
+    {
+        $action = $this->asClient()->get('v3/actions', [
+            'filter' => ['callpower_campaign_id' => $callpowerCampaignId],
+        ]);
+
+        if (! $action['data']) {
```
DFurnes: This should be checking if there's an item at index `['data'][0]` or `['data'][0]['id']` right? I believe we'll have a `data` index even if there isn't a matching CallPower campaign… it'll just be empty.

app/Jobs/CreateCallPowerPostInRogue.php
```diff
@@ -0,0 +1,123 @@
...
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+
+class CreateCallPowerPostInRogue implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable;
+
+    /**
+     * The call parameters send from CallPower.
```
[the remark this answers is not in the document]
chloealee: oops thank you!

app/Services/Rogue.php
chloealee: you're totally right, thank you for spotting that! Updated!

composer.lock
```diff
@@ -4024,12 +4142,11 @@
...
             },
             "require-dev": {
-                "symfony/phpunit-bridge": "^3.4 || 4.0"
+                "nyholm/psr7": "^1.1",
+                "symfony/phpunit-bridge": "^3.4 || ^4.0"
             },
             "suggest": {
-                "psr/http-factory-implementation": "To use the PSR-17 factory",
-                "psr/http-message-implementation": "To use the HttpFoundation factory",
-                "zendframework/zend-diactoros": "To use the Zend Diactoros factory"
```
aaronschachter: Refs https://dosomething.slack.com/archives/CA0N5FXND/p1554214878004800?thread_ts=1554213885.001000&cid=CA0N5FXND

PR #65, opened 2019-03-13
Send password reset emails for new RTV users

aaronschachter:
#### What's this PR do?

* Sends an "Activate Account" email to new Rock The Vote users who have opted-in to email communication.

* Fixes setting Customer.io profile to `unsubscribed` if new user has not opt-ed in to email

* Modifies the RTV import to process all records, so we can test on QA.  I've modified our Call to Action Email event triggered campaign in Customer.io Staging to only send for users with `dosomething` domains, so we can test imports on QA without sending real emails to users.

While testing on Customer.io staging, confirm email is sent for new accounts with a `dosomething.org` domain, and email is not sent for non DS domains. I've been testing by uploading a CSV with my DS email and personal gmail, prefixed with `+[timestamp]`, e.g. `aschachter+201903121423@dosomething.org` so I can test that emails are sent.

For Customer.io production, emails will be sent for all new users (regardless of domain). We'll also need to modify some other event triggered campaigns in our production Customer.io instance per the [Deployment steps in our spec](https://docs.google.com/document/d/1RbWAAJA-zTfQYv6dBVIcpJqegKJUM0dH4DkVpikRpE0/edit?ts=5c6f0bd1#heading=h.f68k8ccmho5).

* Merging this is blocked until we get final email copy added into Northstar for the `rock-the-vote-activate-account` password reset type.

### DIFF
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -188,11 +188,6 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
      */
     public function handle(Rogue $rogue)
     {
-        if (is_test_email($this->record->email)) {
-            info('progress_log: Skipping test: ' . $this->record->email);
-            return;
-        }
-
         info('progress_log: Processing: ' . $this->record->email);
 
         $user = $this->getUser($this->record);
@@ -200,6 +195,7 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
             $this->updateUser($user, ['voter_registration_status' => $this->record->voter_registration_status]);
         } else {
             $user = $this->createUser($this->record);
+            $this->sendUserPasswordReset($user);
         }
 
         $existingPosts = $rogue->getPost([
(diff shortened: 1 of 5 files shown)


app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -272,11 +270,11 @@ private function createUser($record)
             $userData[$key] = $record->{$key};
         }
     
-        if (!empty($record->email_subscription_status)) {
+        if (isset($record->email_subscription_status)) {
```
katiecrane: Won't `isset` be `true` even if `$record->email_subscription_status` is `false`?

aaronschachter: Yup! Our `$record->email_subscription_status` has cast the user's email opt-in preferences to a boolean here, so my understanding is that we'd want to update Northstar with `true` or `false` for these new users. We use this to determine whether to send the Activate Account welcome email.

katiecrane: OH wow okay I reversed the red and green lines in my head and that's why I was confused! I'm on board with this!

PR #66, opened 2019-03-20
Config to disable sending RTV emails

aaronschachter:
#### What's this PR do?

Adds support for a `ROCK_THE_VOTE_RESET_ENABLED` environment variable to disable sending RTV Activate Account emails. There's still configuration work needed within the Customer.io UI to avoid sending double emails to new RTV users -- this config var allows us to toggle when to start sending `call_to_action_email` events with type `rock-the-vote-activate-account` in Customer.io.

👀 

Don't want to be blocked from deploying #65 to prod per pending Customer.io config changes.

### DIFF
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -330,13 +330,25 @@ class CreateRockTheVotePostInRogue implements ShouldQueue
      */
     private function sendUserPasswordResetIfSubscribed($user)
     {
+        /**
+         * Our Customer.io event triggered campaign that sends these RTV password resets should be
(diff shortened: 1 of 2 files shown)


app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -330,13 +330,25 @@ private function updateUser($user, $data)
      */
     private function sendUserPasswordResetIfSubscribed($user)
     {
+        /**
+         * Our Customer.io event triggered campaign that sends these RTV password resets should be
+         * configured to not send the email to an unsubscribed user, but let's sanity check anyway.
+         */
         if (!$user->email_subscription_status) {
```
chloealee: love this safety check!

PR #67, opened 2019-03-21
Display Rock The Vote config

aaronschachter:
#### What's this PR do?

Renders the variables in our Rock The Vote import config, to make it easy to determine whether the Send Email config introduced in #66 is enabled.

Also hides the import forms for jobs we no longer run (and plan to ✂️ from codebase in a subsequent PR for spring cleaning 🍃)

👀 

Beats checking the Heroku settings?

### DIFF
--- a/resources/views/pages/import.blade.php
+++ b/resources/views/pages/import.blade.php
@@ -4,13 +4,18 @@
 
 @if (in_array(request()->input('type'), \Chompy\ImportType::all()))
     <div>
+        @if (request()->input('type') === \Chompy\ImportType::$rockTheVote)
+          <p>
+            Send Activate Account email: <strong>{{ config('import.rock_the_vote.reset.enabled') ? 'Enabled' : 'Disabled' }}</strong>
+          </p>
+        @endif
         <form action={{url('/import')}} method="post" enctype="multipart/form-data">
             {{ csrf_field() }}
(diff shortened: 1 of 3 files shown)


resources/views/pages/import.blade.php
```diff
@@ -2,23 +2,37 @@
...
 
-@if (in_array(request()->input('type'), \Chompy\ImportType::all()))
+@if ($type === \Chompy\ImportType::$rockTheVote)
     <div>
+        <h1>Rock The Vote</h1>
+        <p>Creates or updates users and posts from CSV.</p>
+        <h4>Users</h4>
+        <dl>
+            <dt>Source:</dt><dd>{{ config('services.northstar.client_credentials.client_id') }}</dd>
+            <dt>Reset email enabled</dt><dd>{{ $resetConfig['enabled'] ? 'true' : 'false'}}</dd>
```
chloealee: Does this have to be a ternary since `$resetConfig['enabled']` is a boolean? Can we just return `$resetConfig['enabled']`?

aaronschachter: Yeah, when I tried printing the raw boolean value - it rendered as an empty string when `false`.

chloealee: oh interesting!

PR #68, opened 2019-03-22
Set email_subscription_topics for new users

aaronschachter:
#### What's this PR do?

* Subscribes new users to the community newsletter per https://www.pivotaltracker.com/n/projects/2019429. Refs https://github.com/DoSomething/northstar/pull/851 (renaming `action` to `community`)

* Adds a `get_user_source` helper to DRY setting a new user.source across all import jobs (refs #64)

Verify new users have `email_subscription_status` set to `['community']`

Feeling just about ready to  ✂️🔥 the import jobs we no longer use (TurboVote, FB Share) for epic spring cleaning 🍃 

#### Relevant tickets

Fixes https://www.pivotaltracker.com/n/projects/2019429

### DIFF
--- a/app/Jobs/CreateRockTheVotePostInRogue.php
+++ b/app/Jobs/CreateRockTheVotePostInRogue.php
@@ -23,11 +23,14 @@ class RockTheVoteRecord {
         $this->first_name = $record['First name'];
         $this->last_name = $record['Last name'];
         $this->mobile = $record['Phone'];
-        $this->source = config('services.northstar.client_credentials.client_id');
+        $this->source = get_user_source();
 
         $emailOptIn = $record['Opt-in to Partner email?'];
         if ($emailOptIn) {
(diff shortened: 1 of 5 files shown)


app/Jobs/CreateRockTheVotePostInRogue.php
```diff
@@ -23,11 +23,14 @@ public function __construct($record)
         $this->first_name = $record['First name'];
         $this->last_name = $record['Last name'];
         $this->mobile = $record['Phone'];
-        $this->source = config('services.northstar.client_credentials.client_id');
+        $this->source = get_user_source();
```
DFurnes: We should be able to omit this field & Northstar will automatically set the source based on the OAuth client!

aaronschachter: Oh, I didn't know that! I'll remove this helper then.

PR #72, opened 2019-03-26
Refactor import type as route parameter instead of query parameter

aaronschachter:
#### What's this PR do?

* Changes import by csv route from `import?type=:type` to `import/:type`, to DRY requesting the `type` query parameter throughout the codebase

* Cleans up the `ImportController`:  The changes #67 hardcoded the `show` method to display RTV config, which is fine as we're no longer using/supporting the Turbovote or Facebook Share types, but this PR will make it easier to add more upload CSV forms, should we ever need them

👀 

Next up, removing deprecated code / import types? 🔪 🍖 🍽 

#### Relevant tickets

https://www.pivotaltracker.com/story/show/164114650

### DIFF
--- a/resources/views/components/nav.blade.php
+++ b/resources/views/components/nav.blade.php
@@ -13,7 +13,7 @@
         <div id="navbar" class="navbar-collapse collapse">
             <ul class="nav navbar-nav navbar-right">
                 @if (Auth::user())
-                    <li class="{{ request()->input('type') === \Chompy\ImportType::$rockTheVote ? 'active' : '' }}">
+                    <li class="{{ request()->url('') === 'import/'.\Chompy\ImportType::$rockTheVote ? 'active' : '' }}">
                         <a class="nav-item nav-link" href="/import?type={{ \Chompy\ImportType::$rockTheVote }}">
                             Rock The Vote
                         </a>
--- a/resources/views/components/nav.blade.php
+++ b/resources/views/components/nav.blade.php
@@ -8,13 +8,13 @@
                 <span class="icon-bar"></span>
             </button>
 
-            <a class="navbar-brand" href="/">Chompy</a>
+            <a class="navbar-brand">Chompy</a>
         </div>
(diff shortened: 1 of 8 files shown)


resources/views/pages/import.blade.php
```diff
@@ -2,47 +2,34 @@
...
-                        </span>
-                    </label>
-                    <input type="hidden" name="import-type" value={{ app('request')->input('type') }}>
-                    <input type="text" class="form-control" readonly>
-                </div>
+<div>
+    @if ($type === \Chompy\ImportType::$rockTheVote)
+    @include('pages.partials.rock-the-vote', ['config' => $config])
+    @endif
+    <form action={{ app('request')->url() }} method="post" enctype="multipart/form-data">
```
DFurnes: This should probably be explicitly pointed at the "store" route with `route('import.store')`

resources/views/components/nav.blade.php
```diff
@@ -8,13 +8,13 @@
...
 
-            <a class="navbar-brand" href="/">Chompy</a>
+            <a class="navbar-brand">Chompy</a>
         </div>
         <div id="navbar" class="navbar-collapse collapse">
             <ul class="nav navbar-nav navbar-right">
                 @if (Auth::user())
-                    <li class="{{ request()->input('type') === \Chompy\ImportType::$rockTheVote ? 'active' : '' }}">
-                        <a class="nav-item nav-link" href="/import?type={{ \Chompy\ImportType::$rockTheVote }}">
+                    <li class="{{ str_contains(request()->url(), $rockTheVote) ? 'active' : '' }}">
```
DFurnes: Doing a string comparison on the whole URL feels like it could result in subtle bugs (e.g. if a future import is a substring of another one, or shows up in an unrelated query-string). Let's try matching on the path instead:

```blade
<li class="{{ request()->is('import/' . $rockTheVote) ? 'active' : '' }}">
```

PR #73, opened 2019-03-26
Unit tests

chloealee:
#### What's this PR do?
Adds unit tests outlined in the "Testing" section of the [CallPower Tech Spec.](https://docs.google.com/document/d/1aDNLKZ6BPto65Ce5RDJkXf8LpKGIFdftO13WgZgkO48/edit)

Does this logic make sense? I think I wrote tests for all the scenarios that we could test for from the doc - am I missing any? 

#### Relevant tickets
Adds unit tests for #64

### DIFF
--- a/tests/Unit/CallPowerTest.php
+++ b/tests/Unit/CallPowerTest.php
@@ -11,25 +11,6 @@ use DoSomething\Gateway\Resources\NorthstarUser;
 class CallPowerTest extends TestCase
 {
     /**
-     * Mock the getUser Northstar Call.
-     *
-     * @return user
-     */
-    public function mockGetUserNSCall()
-    {
-        $this->northstarMock->shouldReceive('getUser')->andReturnUsing(function ($type, $id) {
-            return new NorthstarUser([
-                    'id' => $type === 'id' ? $id : $this->faker->northstar_id,
-                    'first_name' => $this->faker->firstName,
-                    'last_name' => $this->faker->lastName,
-                    'birthdate' => $this->faker->date,
-                    'email' => $this->faker->email,
-                    'mobile' => $this->faker->phoneNumber,
-                ]);
-        });
-    }
-
-    /**
      * Test that a post with a completed status successfully sends to Rogue.
      *
      * @return void
@@ -95,7 +76,7 @@ class CallPowerTest extends TestCase
     public function testFailedPostWithInvalidCallPowerCampaignId()
     {
         // Mock the Northstar call.
-        $this->mockGetUserNSCall();
+        $this->mockGetNorthstarUser();
 
         // If the provided Action ID doesn't exist in Rogue, this method throws:
         $this->rogueMock->shouldReceive('getActionIdFromCallPowerCampaignId')->andThrow(new Exception(500));
@@ -127,7 +108,7 @@ class CallPowerTest extends TestCase
     public function testFailedPostCreationInRogue()
     {
         // Mock the Northstar call.
-        $this->mockGetUserNSCall();
+        $this->mockGetNorthstarUser();
 
         // Mock the call to Rogue to get the action id.
         $this->rogueMock->shouldReceive('getActionIdFromCallPowerCampaignId');
--- a/tests/Unit/CallPowerTest.php
+++ b/tests/Unit/CallPowerTest.php
@@ -4,9 +4,8 @@ namespace Tests\Unit;
 
 use Exception;
 use Tests\TestCase;
-use Illuminate\Support\Facades\Bus;
+use Illuminate\Http\Response;
 use Chompy\Jobs\CreateCallPowerPostInRogue;
-use DoSomething\Gateway\Resources\NorthstarUser;
 
 class CallPowerTest extends TestCase
 {
@@ -17,8 +16,12 @@ class CallPowerTest extends TestCase
      */
     public function testCompletedCallStatus()
     {
-        // Fake the job being dispatched.
-        Bus::fake();
+        // Mock the Northstar call.
+        $this->mockGetNorthstarUser();
+
+        // Mock the Rogue calls.
+        $this->rogueMock->shouldReceive('getActionIdFromCallPowerCampaignId');
+        $this->rogueMock->shouldReceive('createPost');
 
         // Create a post with completed as the status.
         $response = $this->withChompyApiKey()->json('POST', 'api/v1/callpower/call', [
@@ -35,8 +38,6 @@ class CallPowerTest extends TestCase
         ]);
 
         $response->assertSuccessful();
-
-        Bus::assertDispatched(CreateCallPowerPostInRogue::class);
     }
 
     /**
@@ -46,8 +47,12 @@ class CallPowerTest extends TestCase
      */
     public function testBusyCallStatus()
     {
-        // Fake the job being dispatched.
-        Bus::fake();
+        // Mock the Northstar call.
+        $this->mockGetNorthstarUser();
+
+        // Mock the Rogue calls.
+        $this->rogueMock->shouldReceive('getActionIdFromCallPowerCampaignId');
+        $this->rogueMock->shouldReceive('createPost');
 
         // Create a post with completed as the status.
         $response = $this->withChompyApiKey()->json('POST', 'api/v1/callpower/call', [
@@ -64,8 +69,6 @@ class CallPowerTest extends TestCase
         ]);
 
         $response->assertSuccessful();
-
-        Bus::assertDispatched(CreateCallPowerPostInRogue::class);
     }
 
     /**
@@ -83,38 +86,6 @@ class CallPowerTest extends TestCase
 
         // Since we don't have an action in Rogue, we shouldn't make a post:
         $this->rogueMock->shouldNotReceive('createPost');
-
-        $this->expectException(Exception::class);
-
-        CreateCallPowerPostInRogue::dispatch([
-            'mobile' => '1234567891',
-            'callpower_campaign_id' => 1,
-            'status' => 'busy',
-            'call_timestamp' => '2017-11-09 06:34:01.185035',
-            'call_duration' => 50,
-            'campaign_target_name' => 'Mickey Mouse',
-            'campaign_target_title' => 'Representative',
-            'campaign_target_district' => 'FL-7',
-            'callpower_campaign_name' => 'Test',
-            'number_dialed_into' => '+12028519273',
-        ]);
-    }
-
-    /**
-     * Test a failed record if the post is not able to be created.
-     *
-     * @return void
-     */
-    public function testFailedPostCreationInRogue()
-    {
-        // Mock the Northstar call.
-        $this->mockGetNorthstarUser();
-
-        // Mock the call to Rogue to get the action id.
-        $this->rogueMock->shouldReceive('getActionIdFromCallPowerCampaignId');
-
-        // Mock a failed call to Rogue when a post is not created.
-        $this->rogueMock->shouldReceive('createPost')->andThrow(new Exception(500));
 
         $this->expectException(Exception::class);
 
--- a/tests/Unit/CallPowerTest.php
+++ b/tests/Unit/CallPowerTest.php
@@ -1,0 +1,184 @@
+<?php
+
+namespace Tests\Unit;
+
+use Exception;
+use Tests\TestCase;
+use Illuminate\Support\Facades\Bus;
+use Chompy\Jobs\CreateCallPowerPostInRogue;
+use DoSomething\Gateway\Resources\NorthstarUser;
+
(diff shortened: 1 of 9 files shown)


tests/WithMocks.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        // Reset mocked time, if set.
+        Carbon::setTestNow(null);
+
+        // Fake the storage driver.
+        Storage::fake('public');
+
+        // Get a new Faker generator from Laravel.
+        $this->faker = app(\Faker\Generator::class);
+        $this->faker->addProvider(new \FakerNorthstarId($this->faker));
```
chloealee: hi @DFurnes ! I was using Rogue as an example to set up mocks for tests. When I run this first test however, I get 
```
There was 1 error:

1) Tests\Unit\CallPowerTest::testFailedPostWithInvalidCallPowerCampaignId
Error: Class 'FakerNorthstarId' not found

/home/vagrant/DoSomething.org/chompy/tests/WithMocks.php:33
/home/vagrant/DoSomething.org/chompy/tests/TestCase.php:28
```
and I'm not sure why? Is there anything I have to do that I'm missing so it can register this line? I tried to put it in a `use` statement at the top of the file but that didn't work either...

DFurnes: @chloealee It's saying that the class isn't found because that's a file from Rogue! Each application is its own self-contained "package", and can't directly reference code from other codebases (that's why we expose APIs).

Here's [the file that's referencing](https://github.com/DoSomething/rogue/blob/master/database/faker/FakerNorthstarId.php)  in Rogue, which you'd need to copy over to use here.

chloealee: thanks, @DFurnes ! I thought I did that with [this file](database/faker/FakerNorthstarId.php) though in this PR?

chloealee: Oh I just needed to do a `composer update`!

tests/Unit/CallPowerTest.php
```diff
@@ -0,0 +1,91 @@
...
+            'call_duration' => 50,
+            'campaign_target_name' => 'Mickey Mouse',
+            'campaign_target_title' => 'Representative',
+            'campaign_target_district' => 'FL-7',
+            'callpower_campaign_name' => 'Test',
+            'number_dialed_into' => '+12028519273',
+        ]);
+
+        $response->assertStatus(200);
+    }
```
chloealee: @DFurnes still trying to wrap my head around what we are testing here! Is this on the right path?

tests/Unit/CallPowerTest.php
```diff
@@ -0,0 +1,92 @@
...
+     * Test that a post with an invalid CallPower campaign id is not made,
+     * and enters into the failed_jobs table.
+     *
+     * @return void
+     */
+    public function testFailedPostWithInvalidCallPowerCampaignId()
+    {
+        // Mock the Rogue API call to get the action id.
+        $this->mock(Rogue::class)
+            ->shouldNotReceive('getActionIdFromCallPowerCampaignId')->andReturn(new Exception(500))
```
chloealee: I also don't think this is the way to test failed jobs the more I think about it and the more research I do... should this be testing if it makes it into the queued job instead? Maybe this test (to make sure it makes it in the `failed_jobs` table) is not meant for unit test functionality since it is more of an overall "feature" and too big for this type of testing?

tests/WithMocks.php
DFurnes: Ahhh sorry! Was looking too fast! 🤦‍♂️

tests/Unit/CallPowerTest.php
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Test that a post with an invalid CallPower campaign id is not made,
+     * and enters into the failed_jobs table.
+     *
+     * @return void
+     */
+    public function testFailedPostWithInvalidCallPowerCampaignId()
+    {
+        // Mock the Rogue API call to get the action id.
+        $this->mock(Rogue::class)
```
DFurnes: I think this totally is on the right track! These tests can probably be whittled down to the job itself (since we test the validation and authorization components above). So, maybe something like:

```js
// If the provided Action ID doesn't exist in Rogue, this method throws:
$this->rogueMock->shouldReceive('getActionIdFromCallPowerCampaignId')->andThrow(new Exception(500));

// Since we don't have an action in Rogue, we shouldn't make a post:
$this->rogueMock->shouldNotReceive('createPost');

CreateCallPowerPostInRogue::dispatch([
  // ...
]);
```

tests/WithMocks.php
```diff
@@ -0,0 +1,71 @@
...
+                    'last_name' => $this->faker->lastName,
+                    'birthdate' => $this->faker->date,
+                    'email' => $this->faker->email,
+                    'mobile' => $this->faker->phoneNumber,
+                ]);
+        });
+
+        // Rogue Mock
+        $this->rogueMock = $this->mock(Rogue::class);
+        $this->rogueMock->shouldReceive('getActionIdFromCallPowerCampaignId')->andReturn($this->faker->randomDigitNotNull);
```
DFurnes: We probably don't want to configure these mock expectations globally, since we'll want different responses in each test case. (Especially `createPost` below, which we'll want to configure in each test so we can assert the correct post was made!)

tests/Unit/CallPowerTest.php
DFurnes: Yeah, I agree! I think we can trust that Laravel handles putting failed jobs into the `failed_jobs` table since that's built in to the framework. We're just interested in whether the job succeeds or fails for the purposes of this test.

chloealee: ok awesome, thanks! I'll remove this :)

tests/Unit/CallPowerTest.php
```diff
@@ -0,0 +1,92 @@
...
+            'call_timestamp' => '2017-11-09 06:34:01.185035',
+            'call_duration' => 50,
+            'campaign_target_name' => 'Mickey Mouse',
+            'campaign_target_title' => 'Representative',
+            'campaign_target_district' => 'FL-7',
+            'callpower_campaign_name' => 'Test',
+            'number_dialed_into' => '+12028519273',
+        ]);
+
+        $response->assertStatus(200);
```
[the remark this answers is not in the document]
chloealee: ooo I like that, thank you @DFurnes ! Updated :)

tests/Unit/CallPowerTest.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Test that a post with an invalid CallPower campaign id is not made.
+     *
+     * @return void
+     */
+    public function testFailedPostWithInvalidCallPowerCampaignId()
+    {
+        // If the provided Action ID doesn't exist in Rogue, this method throws:
+        $this->rogueMock->shouldReceive('getActionIdFromCallPowerCampaignId')->andThrow(new Exception(500));
```
chloealee: @DFurnes I'm a little stuck on this one because it does throw an exception, as we want it to, but then the whole test errors out: 
```
There was 1 error:

1) Tests\Unit\CallPowerTest::testFailedPostWithInvalidCallPowerCampaignId
Exception: 500
```
How do we get around this to say we want this to happen? I added `$this->expectException(Exception::class);` at the bottom to assert this but the test errors out before it gets to that point I think.

chloealee: I _think_ I figured it out to move the `$this->expectException(Exception::class);` to before the job is called so it doesn't error out?

DFurnes: Yes, that sounds right to me! That should be telling PHPUnit to "expect" that type of exception to be thrown, and so it swallows it up (and I _think_ actually fails the test if it _isn't_ thrown). 👍

tests/Unit/CallPowerTest.php
```diff
@@ -0,0 +1,185 @@
...
+    public function testFailedUserCreationInNorthstar()
+    {
+        // Mock a failed Northstar response.
+        $this->northstarMock->shouldReceive('getUser')->andReturn(null);
+        $this->northstarMock->shouldReceive('createUser')->andThrow(new Exception(500));
+
+        // Since we don't have a user, these methods should not be hit.
+        $this->rogueMock
+            ->shouldNotReceive('getActionIdFromCallPowerCampaignId');
+        //     ->shouldNotReceive('createPost');
```
katiecrane: Should this still be commented out?

katiecrane:
@chloealee I think this is looking really awesome! There are a few `.DS_Store` files that I don't think should be included though

tests/Unit/CallPowerTest.php
chloealee: thanks for catching that! Added in :)

chloealee:
> @chloealee I think this is looking really awesome! There are a few .DS_Store files that I don't think should be included though


@katiecrane yikes! I'm not sure how that got in there. I'm trying to delete it but don't even see it in my files on my local so can't do that. Do you know how to get rid of it?

katiecrane:
@chloealee Rogue has [this in the .gitignore](https://github.com/DoSomething/rogue/blob/master/.gitignore#L10) which Chompy does not, maybe that'll do it?

DFurnes:
Yeah, we should definitely add `.DS_Store` to this app's gitignore. You can remove any existing ones using the Terminal with `rm` (since they're hidden files they won't show up in the Finder).

chloealee:
ooof this got way out of wack and the rebasing got really messy and looks like I accidentally removed all the tests from this PR :( I'll ping when I've added back in.

chloealee:
@DFurnes this should be back in order and ready for final review!

tests/Unit/CallPowerTest.php
```diff
@@ -0,0 +1,184 @@
...
+use DoSomething\Gateway\Resources\NorthstarUser;
+
+class CallPowerTest extends TestCase
+{
+    /**
+     * Mock the getUser Northstar Call.
+     *
+     * @return user
+     */
+    public function mockGetUserNSCall()
```
DFurnes: Let's move this into the `WithMocks` trait, since it seems like something we'd likely want to use in other tests. And this is super nitpicky, but I'd love to rename it to `mockGetNorthstarUser` rather than abbreviating `NS`! 

🤓

tests/Unit/CallPowerTest.php
```diff
@@ -0,0 +1,184 @@
...
+        // Mock the Northstar call.
+        $this->mockGetUserNSCall();
+
+        // Mock the call to Rogue to get the action id.
+        $this->rogueMock->shouldReceive('getActionIdFromCallPowerCampaignId');
+
+        // Mock a failed call to Rogue when a post is not created.
+        $this->rogueMock->shouldReceive('createPost')->andThrow(new Exception(500));
+
+        $this->expectException(Exception::class);
```
DFurnes: This test doesn't effectively do all that much (since it just asserts that we got the exception we mocked above), so it maybe can be removed. The other "failure" tests seem more helpful, since they're asserting that "if this step fails, it shouldn't try to do the other thing".

tests/Unit/CallPowerTest.php
```diff
@@ -0,0 +1,184 @@
...
+            'campaign_target_name' => 'Mickey Mouse',
+            'campaign_target_title' => 'Representative',
+            'campaign_target_district' => 'FL-7',
+            'callpower_campaign_name' => 'Test',
+            'number_dialed_into' => '+12028519273',
+        ]);
+
+        $response->assertSuccessful();
+
+        Bus::assertDispatched(CreateCallPowerPostInRogue::class);
```
DFurnes: It miiiiiight actually make sense to remove the `Bus::fake()` call from this test (or create another "successful" test down below that directly dispatches the job like the failure ones do) so that we can test the "internals" of the job function as expected (for example, that this payload results in a particular set of Northstar & Rogue API calls).

tests/Unit/CallPowerTest.php
katiecrane: +1 to moving this into `WithMocks`!

tests/Unit/CallPowerTest.php
chloealee: @DFurnes ok! If I remove the `Bus::fake()`, I would want to mock the NS and Rogue calls, right? But if I mock a successful Rogue call, would that be testing anything ?

DFurnes: We'd be asserting the format of the user & post that gets created for the given CallPower payload. So, for example, that the given `call_power_campaign_id` results in a post with the right `action_id` getting created, etc.

chloealee: ah ok sorry I know you've explained this to me so many times! Updated this and the one below to not use `Bus::fake()` anymore!

PR #76, opened 2019-03-27
sends sms_status stop when creating the user

chloealee:
#### What's this PR do?
Sends sms_status stop when creating the user to double check that we aren't opting a user into SMS messaging when creating a new member in CallPower. 

👀 

More details in Slack convo [here](https://dosomething.slack.com/archives/C03V8B1HA/p1553693654110000) and [here](https://dosomething.slack.com/archives/CA0N5FXND/p1553697085009900).

#### Relevant tickets
Suppresses messaging in the endpoint created in #64

### DIFF
--- a/app/Jobs/CreateCallPowerPostInRogue.php
+++ b/app/Jobs/CreateCallPowerPostInRogue.php
@@ -81,6 +81,7 @@ class CreateCallPowerPostInRogue implements ShouldQueue
         if (is_null($user)) {
             $user = gateway('northstar')->asClient()->createUser([
                 'mobile' => $mobile,
+                'sms_status' => 'stop',
             ]);
         }
 


aaronschachter:
I'd think we'd only send the welcome text if new user is created with `active`, but setting status to `stop` will prevent them from possibly getting broadcasts they haven't specifically opted into.

app/Jobs/CreateCallPowerPostInRogue.php
```diff
@@ -81,6 +81,7 @@ private function getOrCreateUser($mobile)
         if (is_null($user)) {
             $user = gateway('northstar')->asClient()->createUser([
                 'mobile' => $mobile,
+                'sms_status' => 'stop',
```
aaronschachter: Nothing needed for merge, but It'd be nice to add all of these constants like this one, and the post source/details as a new array items in the [config](https://github.com/DoSomething/chompy/blob/master/config/import.php), e.g. `config('call_power.user_sms_status`)`, `config('call_power.post.source')`, `type`, etc

chloealee: Love that idea, @aaronschachter ! Made a card for that [here](https://www.pivotaltracker.com/n/projects/2019429/stories/164932452).

chloealee:
> I'd think we'd only send the welcome text if new user is created with active, but setting status to stop will prevent them from possibly getting broadcasts they haven't specifically opted into.

@aaronschachter I think this is ok in this case since we'd want them to opt-in to messaging anyway down the road! Is that correct, @ngjo ?

ngjo:
Yup, that's what we'd want to do! Not send the welcome message and ensure they're `stop` ped for sms messages. Perhaps some cross channel promotion could get them to opt in (if we have their emails), but otherwise we wouldn't text them to change their status since they haven't explicitly opted in.

chloealee:
awesome, thanks for confirming, @ngjo ! I'll merge this in then :)

PR #78, opened 2019-03-29
Removes unused TurboVote and FacebookShare imports

aaronschachter:
#### What's this PR do?

We haven't used this code for months -- if we ever need to add them back, this is the PR to look for!

Verify Rock The Vote import works as expected.

There's a lot of cleanup these imports could use, if we were to keep them around. I'd prefer to remove so we're not copy/pasting any bad habits.

#### Relevant tickets

Fixes https://www.pivotaltracker.com/story/show/164114650


app/Jobs/ImportFacebookSharePosts.php
```diff
@@ -1,122 +0,0 @@
...
-    /**
-     * Create a new job instance.
-     *
-     * @return void
-     */
-    public function __construct($filepath)
-    {
-        $this->filepath = $filepath;
-
-        $this->stats = $this->statsInit();
```
aaronschachter: Seems this is the only job that was posting anything to Stathat (and that calls the `statsInit ` function defined over in the [`ImportToRogue` trait](https://github.com/DoSomething/chompy/blob/0569eb68780d47d2e4e49b7cdcb46e5356599614/app/Traits/ImportToRogue.php#L16)

PR #82, opened 2019-04-01
Removes unused stats and password_resets tables

aaronschachter:
#### What's this PR do?

Adds migration to drop unused tables. Per https://github.com/DoSomething/chompy/pull/46#discussion_r217419461 - it looks like adding stats for large files isn't in the works any time soon, so let's continue our spring cleaning by removing code we don't need.

The `password_resets` table isn't used either. 

Do imports still work? They should.

I think our documentation could use some TLC too. I started that here, but held off on more changes to keep them separate from dropping deprecated tables.


database/migrations/2019_04_01_000000_drop_password_resets_and_stats_tables.php
```diff
@@ -0,0 +1,18 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::dropIfExists('password_resets');
+        Schema::dropIfExists('stats');
+    }
+}
```
katiecrane: This migration needs a [corresponding `down()` function to reverse what happens in the `up()`](https://laravel.com/docs/5.6/migrations#migration-structure) so that it doesn't error out or get confused should we try to rollback any migrations. I think you should be able to grab exactly what is in the `up()` method of the original [`password_resets`](https://github.com/DoSomething/chompy/blob/master/database/migrations/2014_10_12_100000_create_password_resets_table.php#L16-L20) and migration, but for stats I think you'll have to combine [the original](https://github.com/DoSomething/chompy/blob/master/database/migrations/2018_04_19_144115_add_stats_table.php#L16-L21) and [another one that added a column](https://github.com/DoSomething/chompy/blob/master/database/migrations/2018_04_23_205823_add_filename_to_stats.php#L16-L18).

aaronschachter: Thanks, updated!

docs/contributing.md
```diff
@@ -51,6 +27,6 @@ There are three different pusher app environments and the credentials for connec
 
 The [Debug Console](https://dashboard.pusher.com/apps/550921/console/realtime_messages) in the Pusher Dashboard will let you see events coming in and going out of each pusher app.
```
aaronschachter: Per [Slack thread](https://dosomething.slack.com/archives/CA0N5FXND/p1554143890017100), I couldn't figure out how to get access to see this.

aaronschachter:
Oh sadness - this Heroku build failed on QA:
```
[2019-04-02 20:08:29] production.ERROR: Class 'DropPasswordResetsAndStatsTables' not found {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Class 'DropPasswordResetsAndStatsTables' not found at /app/vendor/laravel/framework/src/Illuminate/Database/Migrations/Migrator.php:418)
[stacktrace]
```

PR #83, opened 2019-04-02
Multi-select email subscription topics

aaronschachter:
#### What's this PR do?

Allows an admin to subscribe users to multiple topics when uploading. Also fixes a typo in the help text 👨‍🎓 

👀 

#### Relevant tickets

This was a nice-to-have reported by @ngjo, let it be had!

### DIFF
--- a/app/Jobs/ImportEmailSubscription.php
+++ b/app/Jobs/ImportEmailSubscription.php
@@ -108,9 +108,11 @@ class ImportEmailSubscription implements ShouldQueue
     private function updateUser($user)
     {
         $existingTopics = ! empty($user->email_subscription_topics) ? $user->email_subscription_topics : [];
+        $newTopics = array_unique(array_merge($existingTopics, $this->email_subscription_topics));
+
         gateway('northstar')->asClient()->updateUser($user->id, [
             'email_subscription_status' => true,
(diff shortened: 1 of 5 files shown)


app/Jobs/ImportEmailSubscription.php
```diff
@@ -108,9 +108,11 @@ private function createUser()
     private function updateUser($user)
     {
         $existingTopics = ! empty($user->email_subscription_topics) ? $user->email_subscription_topics : [];
+        $newTopics = array_unique(array_merge($existingTopics, $this->email_subscription_topics));
```
aaronschachter: @katiecrane I was able to add duplicate subscription topics, this seems like it'd be a useful sanity check to add into Northstar validation?

katiecrane: @aaronschachter that is a great call! right now Northstar assumes it is getting the complete list of subscription topics for a user with no duplicates.

katiecrane: also wow `array_unique`, didn't realize that existed, now there's some other code I can simplify 😈

katiecrane: I made a barebones card for this Northstar issue so we don't forget about it! https://www.pivotaltracker.com/story/show/165074676

PR #88, opened 2019-04-08
adds logging before sending job to create CallPower post and after job is sent

chloealee:
#### What's this PR do?
Adds logging before sending job to create CallPower post and after job is sent. 

Is this enough logging for us to see where the slow down is? Any other places I should put logs to help debug? 

CallPower sent us historical records for our systems to ingest into Rogue. At first, they all ended up in the `failed_jobs` table because there were no action ids for these campaigns in Rogue. When we re-tried the jobs, they started making it into the Rogue QA database (!) however, the queue was working very slowly (we kicked off the retry job on Tuesday and it wasn't done until Friday, with less than 4k records). We hope these logs will help debug why the queue was going so slowly.

### DIFF
--- a/app/Http/Controllers/ThirdParty/CallPowerController.php
+++ b/app/Http/Controllers/ThirdParty/CallPowerController.php
@@ -3,7 +3,6 @@
 namespace Chompy\Http\Controllers\ThirdParty;
 
 use Illuminate\Http\Request;
-use Illuminate\Log\LogManager;
 use Chompy\Http\Controllers\Controller;
 use Chompy\Jobs\CreateCallPowerPostInRogue;
 
--- a/app/Http/Controllers/ThirdParty/CallPowerController.php
+++ b/app/Http/Controllers/ThirdParty/CallPowerController.php
@@ -3,6 +3,7 @@
 namespace Chompy\Http\Controllers\ThirdParty;
 
 use Illuminate\Http\Request;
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/ThirdParty/CallPowerController.php
```diff
@@ -35,7 +35,21 @@ public function store(Request $request)
...
+                'mobile' => $request['mobile'],
+                'callpower_campaign_id' => $request['callpower_campaign_id'],
+                'status' => $request['status'],
+            ])
+        );
+
         // Send to queued job.
         CreateCallPowerPostInRogue::dispatch($parameters);
+
+        Log::debug('sent job to create post with details: ' . json_encode([
```
DFurnes: This will only tell us how long it took for the job to be inserted into the queue (since the jobs are processed later by the queue dyno, not as part of the web request). From my understanding, we had problems processing jobs already in the queue, so this wouldn't tell us much.

From looking at the job, it seems like we actually already have good instrumentation that we can use for debugging! I'd recommend keeping one of these & then we should be good to go! 😸

chloealee: ok sounds good, thanks @DFurnes ! Yeah I wasn't sure how helpful these would be but I will keep the first one!

PR #91, opened 2019-05-16
Update dependencies & remove built files from repo.

DFurnes:
#### What's this PR do?
This pull request updates dependencies & removes the built `public/css` and `public/js` from the repository so we can build these from source instead (using `heroku-postbuild`).

👀 

🔏 

#### Relevant tickets
[#165798645](https://www.pivotaltracker.com/story/show/165798645)


package.json
```diff
@@ -7,16 +7,21 @@
...
     "cross-env": "^5.2.0",
     "jquery": "^3.1.1",
-    "laravel-mix": "^1.0",
+    "laravel-mix": "^4.0",
     "lodash": "^4.17.10",
-    "vue": "^2.1.10"
+    "resolve-url-loader": "^2.3.1",
+    "sass": "^1.20.1",
+    "sass-loader": "^7.1.0",
+    "vue": "^2.1.10",
```
katiecrane: 100% did not realize that Chompy uses vue

DFurnes: I think it might just in there from the default Laravel dependencies, but not 100% sure!

PR #93, opened 2019-05-24
Adds SoftEdge endpoint to create email posts 

chloealee:
#### What's this PR do?
- Adds SoftEdge endpoint to create email posts. 

Can you create a post with this endpoint? 

#### Relevant tickets
Third task in [this Pivotal Card](https://www.pivotaltracker.com/n/projects/2019313/stories/165974086).


chloealee:
I didn't add any tests because unlike CallPower, we don't really manipulate any information or create new users. The only tests I can think of for this is to make sure that everything is required or if a post fails but I don't think those are very useful?

routes/api.php
```diff
@@ -21,4 +21,6 @@
 $router->group(['prefix' => 'v1', 'middleware' => ['third-party-api-token-auth']], function () {
```
chloealee: I was going to create another API token for SoftEdge but then realized that we named the api-token for CallPower something pretty general. I think we have a couple options here: 
- re-name the CallPower API token to something more specific and have CallPower update this on their end so everything continues to work. At the same time, add a separate (and more specifically named) SoftEdge API token so these are different.
- use the same third-party static API key for CallPower and SoftEdge. 
- keep the CallPower api token as is and create a separate (and more specifically named) SoftEdge API token. 

Though it takes work to update on CallPower's end, I think I'm leaning towards the first option so it doesn't get confusing for future devs who hop on this and is more accurately named. Thoughts?

chloealee: oh wait! I think I overthought this. 

I think we can just change the name of `api_key` to `callpower_api_key` and add a `softedge_api_key` and check [here](https://github.com/DoSomething/chompy/blob/master/app/Http/Middleware/AuthenticateWithApiKey.php#L22) if it matches either of those values?

DFurnes: I'm in favor of the first option. I think having different API keys for each third-party's endpoint is the most responsible choice, since that way we're giving each one only the privileges that they needs to do their particular job.

For backwards compatibility, it could make sense to have a property that maps headers to their corresponding "secrets" in this middleware (and we could then easily support a bunch of headers in one `token` middleware), e.g:

```php
protected $keys = [
    'X-DS-Importer-API-Key' => config('app.callpower_key')),
    'X-DS-CallPower-API-Key' => config('app.callpower_key')),
    'X-DS-SoftEdge-API-Key' => config('app.softedge_key')),
];

public function handle($request, Closure $next, ...$headers)
{
    foreach ($headers as $header => $key) {
        if ($request->header($header) == $key) {
            return $next($request);
        }
    }

    throw new AuthenticationException;
}
```

You could then apply the `token` middleware with a list of "acceptable" keys, e.g:

```php
$router->group('middleware' => ['token:X-DS-Importer-API-Key,X-DS-CallPower-API-Key']
```

app/Http/Controllers/ThirdParty/SoftEdgeController.php
```diff
@@ -0,0 +1,44 @@
...
+        $parameters = $request->validate([
+            'northstar_id' => 'required',
+            'action_id' => 'required|integer',
+            'email_timestamp' => 'required|date',
+            'campaign_target_name' => 'required|string',
+            'campaign_target_title' => 'required|string',
+            'campaign_target_district' => 'nullable|string',
+        ]);
+
+        Log::debug('sending job to create post with details: ' . json_encode([
```
DFurnes: This doesn't need to be JSON encoded. The log methods all accept a second array parameter for context:

```php
Log::debug('sending job to create post with details', [
    'northstar_id' => $request['northstar_id'],
    'action_id' => $request['action_id'],
])
```

app/Jobs/CreateSoftEdgePostInRogue.php
```diff
@@ -0,0 +1,77 @@
...
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+
+class CreateSoftEdgePostInRogue implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable;
+
+    /**
+     * The call parameters sent from CallPower.
```
DFurnes: `CallPower` → `SoftEdge`

app/Http/Controllers/ThirdParty/SoftEdgeController.php
```diff
@@ -0,0 +1,44 @@
...
+                'northstar_id' => $request['northstar_id'],
+                'action_id' => $request['action_id'],
+            ])
+        );
+
+        // Send to queued job.
+        CreateSoftEdgePostInRogue::dispatch($parameters);
+    }
```
DFurnes: It might be nice to return a simple "success" message here rather than an empty 200.

```php
CreateSoftEdgePostInRogue::dispatch($parameters);

return response()->json(['success' => true]);
```

app/Jobs/CreateSoftEdgePostInRogue.php
chloealee: oops thank you!

routes/api.php
chloealee: I love this idea, thanks @DFurnes ! Where is it getting the `$headers` variable from? I updated code and when I `dd($headers)`, it still comes in as `X-DS-Importer-API-Key` even though I am passing in `X-DS-SoftEdge-API-Key` in my postman response.

DFurnes: Oops sorry, a typo from trying to code in Markdown! 😆 That should be `$this->keys`.

chloealee: oohh I see - it is only taking the first element in the list of acceptable keys, so only `X-DS-Importer-API-Key` since it is the first one in the list. I think this is the correct syntax though?

chloealee: a-ha! I was missing the `...`

chloealee: @DFurnes one question - with this code, couldn't CallPower use their API key header and token to create a successful request to the SoftEdge endpoint (and vice versa)?

DFurnes: Oh okay okay, I re-read this _again_ & I lied to you in my earlier response! 😅 

We want this to read the `...$headers` being passed in to the middleware (which is the list of keys that can be used for that particular route), so the for-loop _should_ be over `$headers`. However, we just want to loop over the values (rather than `$index => $value`) and then we want to use that to compare each actual request header against the corresponding secret.

I updated my comment above & it's hopefully more accurate now!

chloealee: thanks @DFurnes ! Just updated it :) But I think [this](https://github.com/DoSomething/chompy/pull/93#discussion_r288753593) still applies ? With this logic, I can hit the SoftEdge API endpoint with the correct `X-DS-Importer-API-Key` or `X-DS-CallPower-API-Key` and I don't think we want it to do that?

app/Http/Middleware/AuthenticateWithApiKey.php
```diff
@@ -7,22 +7,36 @@
...
+        // Check the header for authorization.
+        foreach ($headers as $header) {
+            if ($request->header($header) === $this->keys[$header]) {
+                return $next($request);
+            }
         }
 
-        return $next($request);
+        throw new AuthenticationException;
     }
```
chloealee: the indentation is SO weird in this file but without this, the Style CI tests wouldn't pass! I don't know what is wrong with this??

routes/api.php
DFurnes: You'd need to apply the middleware to each route, with its allowed tokens. (Right now, you're applying the same middleware, with _all_ the tokens, to both routes in the group.)

For consistency with the way we apply middleware elsewhere, I'd recommend just applying this in each controller's `__construct`. For example, for `SoftEdgeController`:

```php
/**
 * Create a controller instance.
 *
 * @return void
 */
public function __construct()
{
    $this->middleware('token:X-DS-SoftEdge-API-Key');
}
```

chloealee: ah ok I see, thanks @DFurnes ! This is all working now - tested the CallPower endpoint to make sure it only works with `X-DS-Importer-API-Key` and `X-DS-CallPower-API-Key` and the SoftEdge to make sure it only works `X-DS-SoftEdge-API-Key` and all looks good! Ready for final review!

app/Http/Middleware/AuthenticateWithApiKey.php
chloealee: Ok I think this looks better and is passing!

routes/api.php
chloealee: @DFurnes fixed the StyleCi updates - looks better now and is passing so _now_ this is ready for final review!

chloealee:
ah great catch thank you @DFurnes ! I will make a PR for that now!

chloealee:
https://github.com/DoSomething/infrastructure/pull/188

PR #94, opened 2019-06-05
Removes requirement for campaign_target_title and campaign_target_district

chloealee:
#### What's this PR do?
This PR removes requirement for `campaign_target_title` and `campaign_target_district`. If we would like for a user to send to a custom person (e.g. letter to the editor), these fields will be left blank when sent to our system so we will need to remove these as required fields. 

👀 

#### Relevant tickets
[Pivotal Card
](https://www.pivotaltracker.com/n/projects/2019313/stories/165974086)

### DIFF
--- a/app/Http/Controllers/ThirdParty/SoftEdgeController.php
+++ b/app/Http/Controllers/ThirdParty/SoftEdgeController.php
@@ -29,8 +29,8 @@ class SoftEdgeController extends Controller
             'action_id' => 'required|integer',
             'email_timestamp' => 'required|date',
             'campaign_target_name' => 'required|string',
-            'campaign_target_title' => 'required|string',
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/ThirdParty/SoftEdgeController.php
```diff
@@ -29,8 +29,8 @@ public function store(Request $request)
             'action_id' => 'required|integer',
             'email_timestamp' => 'required|date',
             'campaign_target_name' => 'required|string',
-            'campaign_target_title' => 'required|string',
-            'campaign_target_district' => 'nullable|string',
+            'campaign_target_title' => 'string',
+            'campaign_target_district' => 'string',
```
DFurnes: Do these need to be `nullable` if they're sometimes empty?

PR #95, opened 2019-06-06
if title and district are not set, set to null

chloealee:
#### What's this PR do?
This PR sets `campaign_target_title` and `campaign_target_district` to `null` if they are not sent in the request. 

👀 

#### Relevant tickets

Fixes needed work for #94

### DIFF
--- a/app/Jobs/CreateSoftEdgePostInRogue.php
+++ b/app/Jobs/CreateSoftEdgePostInRogue.php
@@ -70,8 +70,8 @@ class CreateSoftEdgePostInRogue implements ShouldQueue
         return json_encode([
             'email_timestamp' => $email['email_timestamp'],
             'campaign_target_name' => $email['campaign_target_name'],
-            'campaign_target_title' => $email['campaign_target_title'],
-            'campaign_target_district' => $email['campaign_target_district'],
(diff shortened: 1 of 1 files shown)


app/Jobs/CreateSoftEdgePostInRogue.php
```diff
@@ -70,8 +70,8 @@ private function extractDetails($email)
         return json_encode([
             'email_timestamp' => $email['email_timestamp'],
             'campaign_target_name' => $email['campaign_target_name'],
-            'campaign_target_title' => $email['campaign_target_title'],
-            'campaign_target_district' => $email['campaign_target_district'],
+            'campaign_target_title' => $email['campaign_target_title'] ? $email['campaign_target_title'] : null,
+            'campaign_target_district' => $email['campaign_target_district'] ? $email['campaign_target_district'] : null,
```
DFurnes: Shouldn't `$email['campaign_target_title']` already return `null` if that array key isn't set?

chloealee: oh you're right! Ok going to close this. I saw something weird on QA but just tested again and all is working 👍

PR #97, opened 2019-06-18
Imports First Name when Creating/Updating User via Email Subscription

chloealee:
#### What's this PR do?
This PR import first name too if available when creating a new account or updating a user via Chompy email subscription.

This also updates the `/import/email-subscription` page to remind to change the `First Name` column to `first_name` in the CSV: 
Screen Shot 2019-06-18 at 1 52 11 PM

👀 

#### Relevant tickets
[Pivotal Card](https://www.pivotaltracker.com/n/projects/2328687/stories/166608437)

### DIFF
--- a/resources/views/pages/partials/email-subscription.blade.php
+++ b/resources/views/pages/partials/email-subscription.blade.php
@@ -1,7 +1,7 @@
 <div class="form-group">
     <h1>Email subscription</h1>
     <p class="lead">
-      Creates/updates users from uploaded CSV, expecting an <code>email</code> column header.
+      Creates/updates users from uploaded CSV, expecting <code>email</code> and <code>first_name</code> column headers.
     </p>
 </div>
 <h3>Users</h3>
@@ -10,7 +10,7 @@
   <div class="col-sm-9">
     <input type="text" class="form-control" name="source-detail" placeholder="breakdown_opt_in" value="{{ old('source-detail') }}">
     <small class="form-text text-muted">
-      Specify the <code>source_detail</code> for new users that will be created from this upload. 
+      Specify the <code>source_detail</code> for new users that will be created from this upload.
     </small>
   </div>
 </div>
--- a/app/Jobs/ImportEmailSubscription.php
+++ b/app/Jobs/ImportEmailSubscription.php
@@ -19,6 +19,12 @@ class ImportEmailSubscription implements ShouldQueue
      */
     protected $email;
     /**
+     * The first name of the user to subscribe.
+     *
+     * @var array
+     */
+    protected $first_name;
+    /**
      * The source detail if new user.
      *
      * @var string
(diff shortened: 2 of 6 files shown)


resources/views/pages/partials/email-subscription.blade.php
```diff
@@ -1,7 +1,7 @@
 <div class="form-group">
     <h1>Email subscription</h1>
     <p class="lead">
-      Creates/updates users from uploaded CSV, expecting an <code>email</code> column header.
+      Creates/updates users from uploaded CSV, expecting <code>email</code> and <code>first_name</code> column headers.
```
DFurnes: Can we remove the [`display: block` styling for `code` tags](https://github.com/DoSomething/chompy/blob/104977bd33e6a6022177d24f2af17d16f30437e3/resources/assets/sass/app.scss#L31-L33) (and if necessary, apply it with a more specific selector wherever it was necessary)? It makes this kinda awkward to read.

DFurnes: I'm not sure where we use this style override, but it might also make more sense to use a [`pre`](https://getbootstrap.com/docs/3.4/css/#code-block) tag there instead, rather than overloading [the inline code element](https://getbootstrap.com/docs/3.4/css/#code-inline) across the board.

DFurnes:
Also, do we have any tests for this importer? If not, it'd be nice to add some if we have time.

resources/views/pages/partials/email-subscription.blade.php
chloealee: > I'm not sure where we use this style override, but it might also make more sense to use a pre tag there instead, rather than overloading the inline code element across the board.

@DFurnes ah interesting! For this though, I think we only want one word at a time wrapped in the `<code>` tag. `<pre>` looks like it is used when you want multiple lines selected, unless I'm understanding incorrectly?

DFurnes: That's correct!

The elements on this page should be `code` tags _without_ the `display: block` override. If we do want this `display: block` override somewhere else, we should use a `pre` tag for that specific case.

chloealee: ohoh I see what you're saying now! Thanks @DFurnes ! I think this is the only place other than this file that has it and with the `display: block` taken out, looks better to me too!
Screen Shot 2019-06-18 at 2 30 07 PM

chloealee: hmm I removed the code from `app.css` but for some reason it isn't registering the changes when I try to add to GitHub?

chloealee: oh nvm, got it!

chloealee:
Cool, I'll submit another PR for the tests @DFurnes !

PR #98, opened 2019-06-20
Adds test to make sure validation is working on email import

chloealee:
#### What's this PR do?
This PR adds a test to make sure validation is working on the email import. 

👀 

I'm not sure there are any other scenarios we'd want to test with the email import? The only other thing I think we could test is to make sure that a user is created/updated in NS with the `first_name` and `email`, but that seems out of scope for Chompy since that would live in NS?

### DIFF
--- a/app/Http/Middleware/VerifyCsrfToken.php
+++ b/app/Http/Middleware/VerifyCsrfToken.php
@@ -12,6 +12,6 @@ class VerifyCsrfToken extends BaseVerifier
      * @var array
      */
     protected $except = [
-        'import/email-subscription',
+        //
     ];
 }
--- a/app/Http/Middleware/VerifyCsrfToken.php
+++ b/app/Http/Middleware/VerifyCsrfToken.php
@@ -12,6 +12,6 @@ class VerifyCsrfToken extends BaseVerifier
      * @var array
      */
     protected $except = [
-        //
+        'import/email-subscription',
     ];
 }
--- a/tests/Unit/ImporterTest.php
+++ b/tests/Unit/ImporterTest.php
@@ -11,11 +11,11 @@ use Illuminate\Validation\ValidationException;
 class ImporterTest extends TestCase
 {
     /**
-     * Test validation rules for email subscription import.
+     * Test validation error thrown without upload-file for email subscription import.
      *
      * @return void
      */
-    public function testValidationRulesForEmailSubscriptionImport()
+    public function testUploadFileValidationRuleForEmailSubscriptionImport()
     {
         $request = new Request;
         $file = UploadedFile::fake()->create('importer_test.csv');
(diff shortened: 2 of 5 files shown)


tests/Unit/ImporterTest.php
```diff
@@ -3,16 +3,51 @@
...
+            'source-detail' => 'test_opt_in',
+            'topics' => [
+                'community',
+            ],
+        ]);
+
+        $controller = new ImportController();
+
+        $this->expectException(ValidationException::class);
+        $controller->store($request, 'email-subscription');
```
DFurnes: This is an interesting way of doing this! What's the advantage to instantiating a controller & request by hand, rather than using Laravel's `$this->post('<url>, $payload)` test method? Is there any risk you could keep some old payload around by mistake by re-using the same `$request` for each step?

DFurnes:
> I think we could test is to make sure that a user is created/updated in NS with the `first_name` and `email`, but that seems out of scope for Chompy since that would live in NS?

It'd be nice to assert that (given an example CSV or CSV row), it makes the right Northstar API call! But totally agreed that you wouldn't want to test anything beyond the boundary of what Chompy does. ➕

tests/Unit/ImporterTest.php
chloealee: @DFurnes I decided to do it this way (via the controller) since there is no exposed API endpoint to hit this! This method in the controller is triggered when a user uploads a CSV. 

>  Is there any risk you could keep some old payload around by mistake by re-using the same $request for each step?

Great catch! It looks like all the separate requests weren't actually firing off, just the first assertion in the test. I've separated these into three different tests to make sure all of these scenarios are actually being tested.

chloealee:
> It'd be nice to assert that (given an example CSV or CSV row), it makes the right Northstar API call! 

@DFurnes I love this idea! I've been working on this fidgeting around but not sure I'll be able to get done by the EOD. Is it ok to review this card just on validation tests and then create a Pivotal Card for the CSV test to make sure it makes the right NS API call?

DFurnes:
@chloealee Absolutely!

chloealee:
cool, card [here](https://www.pivotaltracker.com/n/projects/2328687/stories/166887416)! This is ready for review :)

tests/Unit/ImporterTest.php
DFurnes: We should still be able to do this with Laravel's built-in HTTP methods, since there _is_ a HTTP endpoint for this (even though it's using web sessions, rather than an API token)! 

For example, this would accomplish the same thing for this test:

```php
$user = \Chompy\User::forceCreate(['role' => 'admin']);
$response = $this->be($user)->postJson('import/email-subscription', [
    'source-detail' => 'test_opt_in',
    'topics' => [
        'community',
    ],
]);

$response->assertStatus(422);
```

DFurnes: That said, definitely not something that needs to happen for merge!

chloealee: ah duh thank you, @DFurnes ! I just updated the tests to have the above since we typically write tests this way in other apps and because I think it is a lot more readable! Ready for another review! (Even though I'm technically not at DS anymore, I can still merge, right? #opensource)!

app/Http/Middleware/VerifyCsrfToken.php
```diff
@@ -12,6 +12,6 @@ class VerifyCsrfToken extends BaseVerifier
      * @var array
      */
     protected $except = [
-        //
+        'import/email-subscription',
```
chloealee: Added this here in order to test with `$this->postJson` test method. I was receieved a `419` status code otherwise and found this suggestion [here](https://stackoverflow.com/questions/46266553/why-laravel-api-return-419-status-code-on-post-and-put-method).

DFurnes:
Made one quick tweak to send the CSRF token with the requests (so we don't need to disable that protection for real traffic on the website, and risk an attacker "hijacking" an email import).

Otherwise, looks perfect and going to merge this on in! Miss you already @chloealee!

chloealee:
@DFurnes ah I love the CSRF token update - a lot more clear when it is in the same file!! Miss you so much already!!!!

PR #101, opened 2019-07-09
Failed job permalink

aaronschachter:
#### What's this PR do?

Adds `/failed-job/:id` links to each row in the failed jobs list, permalinking to the failed job (and displaying the entire exception stack trace)

👀 

Would it make sense to add a delete button to the single view? Is that all we'd need in order to clear the queue of jobs we don't want to retry (e.g. the missing `first_name` column header, or a malformed phone number sent as an input parameter)

### DIFF
--- a/app/Http/Controllers/FailedJobController.php
+++ b/app/Http/Controllers/FailedJobController.php
@@ -18,7 +18,26 @@ class FailedJobController extends Controller
     }
 
     /**
-     * Display a listing of the resource.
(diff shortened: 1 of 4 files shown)


DFurnes:
> Would it make sense to add a delete button to the single view? Is that all we'd need in order to clear the queue of jobs we don't want to retry (e.g. the missing `first_name` column header, or a malformed phone number sent as an input parameter)

I think this makes great sense! 👍

PR #102, opened 2019-07-10
Delete failed job via web

aaronschachter:
#### What's this PR do?

Follow up to #101 -- this PR adds a delete button (with a "Are you sure") confirmation to remove a failed job from the queue.

Also displays parameters for failed SoftEdge jobs.

👀 
I've been testing locally against the QA DB and verifying the failed job is no longer accessible upon submitting the delete form.

This will be especially useful for failed jobs that will never complete no matter how many times they are retried (e.g. missing `email` column header)

### DIFF
--- a/app/Http/Controllers/FailedJobController.php
+++ b/app/Http/Controllers/FailedJobController.php
@@ -62,7 +62,7 @@ class FailedJobController extends Controller
     public function show($id)
     {
         $data = \DB::table('failed_jobs')->where('id', '=', $id)->get();
-        if (!isset($data[0])) {
+        if (! isset($data[0])) {
             abort(404);
         }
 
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/FailedJobController.php
```diff
@@ -30,7 +31,7 @@ protected function addParsedPropertiesToFailedJob($failedJob)
         $failedJob->commandName = $json->data->commandName;
         $failedJob->errorMessage = Str::limit($failedJob->exception, 255);
 
-        if ($failedJob->commandName === 'Chompy\Jobs\CreateCallPowerPostInRogue') {
+        if (Str::contains($failedJob->commandName, 'CallPower') || Str::contains($failedJob->commandName, 'SoftEdge')) {
```
DFurnes: Now that there's a couple of these, it might make sense to create an interface:

```php
interface Inspectable {
    /**
     * Returns the parameters passed to this job.
     *
     * @return array
     */
    public function getParameters();
}
```

```php
class CreateSoftEdgePostInRogue implements ShouldQueue, Inspectable
{
    // ...
```

Then, we can check here whether the failed job's class implements this method:

```php
if (array_has(class_implements($failedJob->commandName), Inspectable::class)) {
   // Now we know that $command will have a getParameters method!
}
```

aaronschachter: I'm into it - where's a good spot for the interface to live? `/App/Jobs/Inspectable.php`?

DFurnes: Yeah, I think that works well. 👍

aaronschachter: I had some trouble getting this to work... will give it a shot in another PR

PR #104, opened 2019-07-12
Optional first_name in email subscription

aaronschachter:
#### What's this PR do?

* Modifies email subscription import to ignore a missing `first_name` column per https://dosomething.slack.com/archives/CA0N5FXND/p1562782007004100, so we'll no longer see failed jobs with `ErrorException: Undefined index: first_name` errors

* Fixes bug in Call Power import to properly return error message when action not found for a CallPower Campaign ID (instead of `ErrorException: Undefined offset: 0 in /app/app/Services/Rogue.php:113`)

* Removes unused `tags` function from `ImportFile`

👀

### DIFF
--- a/app/Jobs/ImportFile.php
+++ b/app/Jobs/ImportFile.php
@@ -96,7 +96,7 @@ class ImportFile implements ShouldQueue
                 ImportRockTheVoteRecord::dispatch($record);
             }
             if ($this->importType === ImportType::$emailSubscription) {
-                ImportEmailSubscription::dispatch($record['email'], $record['first_name'], $this->importOptions['source_detail'], $this->importOptions['email_subscription_topics']);
+                ImportEmailSubscription::dispatch($record, $this->importOptions['source_detail'], $this->importOptions['email_subscription_topics']);
             }
             event(new LogProgress('', 'progress', ($offset / $this->totalRecords) * 100));
         }
--- a/app/Jobs/ImportFile.php
+++ b/app/Jobs/ImportFile.php
@@ -60,17 +60,6 @@ class ImportFile implements ShouldQueue
     }
 
     /**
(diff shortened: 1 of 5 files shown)


app/Jobs/ImportFile.php
```diff
@@ -59,17 +59,6 @@ public function __construct($filepath, $importType, $importOptions)
         $this->importOptions = $importOptions;
     }
 
-    /**
-     * Get the tags that should be assigned to the job.
-     * TODO: Is this used anywhere?
```
aaronschachter: I don't think we [use this](https://github.com/DoSomething/chompy/pull/2#discussion_r178354112).

app/Services/Rogue.php
```diff
@@ -110,9 +110,9 @@ public function getActionIdFromCallPowerCampaignId($callpowerCampaignId)
             'filter' => ['callpower_campaign_id' => $callpowerCampaignId],
         ]);
 
-        if (! $action['data'][0]['id']) {
+        if (empty($action['data'])) {
             info('action id not found for CallPower campaign id: ' . $callpowerCampaignId);
-            throw new Exception(500, 'Unable to get action data for CallPower campaign id : ' . $callpowerCampaignId);
+            abort(500, 'Unable to get action data for CallPower campaign id: ' . $callpowerCampaignId);
```
DFurnes: I think (another) reason we wouldn't have been seeing these errors in the logs was that we were using the wrong signature for the [`Exception`](https://www.php.net/manual/en/class.exception.php) constructor (the first parameter is the message, not the error code).

And this is sorta nitpicky, but I wonder if throwing an old-school Exception is preferable to the `abort` helper here, since `abort` throws a `HttpException` with a status code & this isn't actually running within a web request...

PR #118, opened 2020-02-11
Track changes to RTV documentation

aaronschachter:
### What's this PR do?

This pull request aims to move the RTV documentation out of the [wiki](https://github.com/DoSomething/chompy/wiki/Rock-The-Vote) and into `/docs`, to better track changes. The first commit starts with a copy/paste of the contents on the Rock The Vote import wiki page, and then begins to update and remove stale documentation.

👀, 💬 

Once this is merged, I will delete the Github wiki page.

There's room to trim this down even more -- some of this logic is repeated in the Technical Inventory doc that I've linked to in the Overview section. I'm leaning more towards keeping this documented in our internal Google doc vs open source on the web, as the referral code information is a little bit on the how-the-sausage-gets-made side. 🐴 If we are agreed to go that route, I can remove those sections from these docs and the technical Inventory doc could be the single source of truth.

### Relevant tickets

References [Pivotal #171152803](https://www.pivotaltracker.com/story/show/171152803).

### Checklist

### DIFF
--- a/docs/imports/rock-the-vote.md
+++ b/docs/imports/rock-the-vote.md
@@ -155,13 +155,6 @@ Signup Example:
 - If a user shares their UTM'ed URL with other people, there could be duplicate referral codes but associated with different registrants:
   See a [screenshot](https://cl.ly/0v210N283y2X) of what this data looks like (note: the user depicted in this spreadsheet is fake.)
 
-## Next Steps
-
-1.  Create this logic for RTV
-2.  Start importing!
-3.  Move post voter reg status onto NS profile
-4.  Coordinate w/ Marketing for email (weeklydo only) and sms suppression (no status filled)
-
 ## Open Questions
 
 1. How can we distinguish between TV and RTV import? Do we need to?
--- a/docs/imports/rock-the-vote.md
+++ b/docs/imports/rock-the-vote.md
@@ -4,7 +4,7 @@ We import CSVs from Rock The Vote (RTV) to upsert users and their `voter-reg` po
 
 See [VR Technical Inventory](https://docs.google.com/document/d/1xs2C3DNdD5h1j_abBrGVBNrsrxKvwn2VHDWweIEhvqc/edit?usp=sharing) for more details.
 
-## Status Translation Rules
+## Voter Registration Status
 
 Ultimately, there are 4 `post` statuses we want to capture for `voter-reg` posts for Rock the Vote (**Note** RTV doesn't have a `confirmed` status like TurboVote did):
 
@@ -34,7 +34,7 @@ Because we're pulling some of the columns into the post details, data will then 
 
 For example: If a user has a `confirmed` status already from a TV import, and the RTV file suggests that it should be `uncertain`, do not update.
 
-We’ve established this hierarchy because each time a user interacts with the RTV form a new row is created in the CSV. There are the edge cases when a user is chased to finish their registration that they would be interacting with the same row (thus the "steps"). The hierarchy is the simplest approach to dealing with varying statuses, but we anticipate some edge cases that we may need to deal with as they come up.
+We’ve established this hierarchy because each time a user interacts with the RTV form, a new row is created in the CSV. There are the edge cases when a user is chased to finish their registration that they would be interacting with the same row (thus the "steps"). The hierarchy is the simplest approach to dealing with varying statuses, but we anticipate some edge cases that we may need to deal with as they come up.
 
 Here’s one example:
 
@@ -143,15 +143,14 @@ Signup Example:
 }
 ```
 
-## Other Important Information
+## Notes
 
 - The `details` on the post will have the row of information available for data to use.
 - The `submission_created_at` date is when the importer ran. Details about when the registration was created/updated are in the `source_details`.
 - All of these signups will have a `source` of `importer-client` (this is how messaging is suppressed in C.io)
 - All of these posts have a `type` = `voter-reg`
 - The month that the registration came in is what informs the `action` column (e.g., february-2018-rockthevote)
-- For now, if not available in the referral, we will attribute every`voter-reg` post to the Grab the Mic campaign to maintain a 1:1 relationship. This is what we have done with TurboVote.
-- Quasar will be pushed these posts from Rogue but if Data needs to do a deeper analysis, they will use the raw RTV CSV to do that work.
+- All posts are created for a single Action ID, which is set via config variable. We previously would pass Campaign/Run IDs as query parameters to upsert a  `voter-reg` post for.
 - If a user shares their UTM'ed URL with other people, there could be duplicate referral codes but associated with different registrants:
   See a [screenshot](https://cl.ly/0v210N283y2X) of what this data looks like (note: the user depicted in this spreadsheet is fake.)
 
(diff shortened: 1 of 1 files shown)


docs/imports/rock-the-vote.md
```diff
@@ -0,0 +1,120 @@
...
+- All of these signups will have a `source` of `importer-client` (this is how messaging is suppressed in C.io)
+- In early iterations of the import, the month that the registration came in would inform the `action` column (e.g., february-2018-rockthevote)
+- In early iterations of the import, we would pass Campaign/Run IDs as parameters within the referral code to use when upsert a  `voter-reg` post.
+- If a user shares their UTM'ed URL with other people, there could be duplicate referral codes but associated with different registrants:
+  See a [screenshot](https://cl.ly/0v210N283y2X) of what this data looks like (note: the user depicted in this spreadsheet is fake.)
+
+## Open Questions
+
+1. How can we distinguish between TV and RTV import? Do we need to?
+2. RTV has more birthdays in it, can we use that on the NS profile when we create accounts? (TV didn't provide this)
```
aaronschachter: This sounds a bit like https://www.pivotaltracker.com/story/show/170075407.

docs/imports/rock-the-vote.md
```diff
@@ -0,0 +1,120 @@
...
+3. We've been having some Chompy validation issues w/ the TV import -- is this the good time to tackle those things? If so, is a running list the best way of communicating some of the things we've seen in the referral column? There's one specific referral that's different and it's just `ads` -- these are from Google ads and the only way they could set up tracking that Google likes. Would love to talk through best way to deal with these...if it's something unique!
```
aaronschachter: I'm curious about some of the history here, there's some more context about this in the VR Tech inventory doc... seems related to https://www.pivotaltracker.com/story/show/170504511

docs/imports/rock-the-vote.md
```diff
@@ -0,0 +1,120 @@
...
+- The `submission_created_at` date is when the importer ran. Details about when the registration was created/updated are in the `source_details`.
+- All of these signups will have a `source` of `importer-client` (this is how messaging is suppressed in C.io)
+- In early iterations of the import, the month that the registration came in would inform the `action` column (e.g., february-2018-rockthevote)
+- In early iterations of the import, we would pass Campaign/Run IDs as parameters within the referral code to use when upsert a  `voter-reg` post.
+- If a user shares their UTM'ed URL with other people, there could be duplicate referral codes but associated with different registrants:
+  See a [screenshot](https://cl.ly/0v210N283y2X) of what this data looks like (note: the user depicted in this spreadsheet is fake.)
+
+## Open Questions
+
+1. How can we distinguish between TV and RTV import? Do we need to?
```
aaronschachter: Seems like this is not an issue.

docs/imports/rock-the-vote.md
katiecrane: @aaronschachter I'm unable to follow that link, it says:
```
Sorry, we couldn't find that story
It may be in a project you haven't been invited to yet or it may have been deleted.
```

docs/imports/rock-the-vote.md
katiecrane: same with this link!

docs/imports/rock-the-vote.md
```diff
@@ -0,0 +1,120 @@
...
+- If `status` is `complete` and `finish with state` is `no` --> `register-form`
+- If `status` is `complete` and `finish with state` is `yes` --> `register-OVR`
+- If `status` is any of the `step`s --> `uncertain`
+- If `status` is `rejected` --> `ineligible`
+
+Because we're pulling some of the columns into the post details, data will then be able to know if they, for example, pre-registered or why their registration was ineligible.
+
+### Status Hierarchy
+
+If RTV CSV contains multiple records _for a single user_,  we use the following hierarchy to determine which status should be reported on their Rogue post:
```
katiecrane: This sort of sounds like we only use the hierarchy if the multiple rows come in on the same single CSV file - but really it's not based on if there are two rows for a user in a CSV but if a post already exists in Rogue for the `action_id` right?

docs/imports/rock-the-vote.md
katiecrane: Hmm yeah since we don't use TV anymore I think this is a non-issue but maybe @ngjo has other thoughts?

aaronschachter: We do save `rock-the-vote` to the source detail currently, but ideally we want to change this to a stringified object to track referring user ID if the `referral` parameter is set  (addressed in [#171159899](https://www.pivotaltracker.com/story/show/171159899)). Not sure if we ever set this to `turbovote` for the old import

docs/imports/rock-the-vote.md
ngjo: @aaronschachter yup, that pivotal captures the RTV birthday work. We can distinguish between TV & RTF based on the `source_detail`s, I believe

docs/imports/rock-the-vote.md
ngjo: I don't know if we need to do much more on our end here -- if I remember correctly, this was an issue with a bad process for creating voter reg links and people were using different formats/leaving out critical information. I'm pretty sure we've locked down this process now and the referral column is much cleaner and follows a format that they can more easily split the referral info into separate columns for analysis. In my opinion, we shouldn't be handling splitting that info on our end.

One thought, and I'm not sure if it's helpful, is if it'd be useful for Tej/others doing analysis to store referral information somewhere other than it's being stored now...

docs/imports/rock-the-vote.md
ngjo: Yea, I don't think is an issue anymore, but might be worth connecting with Tej on what the backup would be if RTV went down or anything in case this would be something we'd still want to be able to distinguish.

If anything changes on the `source_detail`s, please loop in Anthony/Tej/Data because Customer.io is looking at `source_detail`s and Tej/data is probably querying off of `source_detail`s

docs/imports/rock-the-vote.md
aaronschachter: per my chat with Tej last week, I think the ideal is to be able to filter voter registrations in Looker by the `source` and `source_detail` we include in a referral code.

docs/imports/rock-the-vote.md
aaronschachter: I left a question in the inventory doc about the TV backup!  Definitely be working slowly and cross-functionally through any new `source_detail` changes we make, if any.

docs/imports/rock-the-vote.md
aaronschachter: Removed this from open questions, as that work is referred to in https://www.pivotaltracker.com/story/show/170504511

aaronschachter:
I think this is in decent shape for adding into the repo as is, and deleting the [existing wiki page](https://github.com/DoSomething/chompy/wiki/Rock-The-Vote). The next round of changes I'd like to make is trimming down some of the details on upserting users and posts and DRY any duplicating info in the Technical Inventory.

aaronschachter:
@katiecrane @ngjo Is this good for merge? I think a next revision/PR would be to trim down some of the wording around the import logic (e.g. creating/updating users)

PR #119, opened 2020-02-25
List imported files

aaronschachter:
### What's this PR do?

This pull request adds a new `ImportFile` model to save a record for each import file uploaded to Chompy, and displays them on a new `/import-files` page.

Screen Shot 2020-02-25 at 10 47 37 AM

Related changes:
* Renames `Controllers/Web/ImportController` to `Controllers/Web/ImportFileController`
* Renames `Jobs/ImportFile` to `Jobs/ImportFileRecords`  

Submit either the RTV or email subscription import form and verify a record is created at `/import-files`

As we're beginning to potentially take on new Chompy work with more 👀 on the web app and codebase, keeping a simple log of each import file created (and later deleted, after all rows are dispatched for import) felt like a quick win. 

There is no plan to backfill this list with old uploads.

### Relevant tickets

References [Pivotal #171472626](https://www.pivotaltracker.com/story/show/171472626).

### Checklist


app/Http/Controllers/ImportFileController.php
```diff
@@ -27,7 +28,7 @@ public function __construct()
      *
      * @param string $importType
      */
-    public function show($importType)
+    public function create($importType)
```
mendelB: What's the reasoning for renaming this action to `create` here?

routes/web.php
```diff
@@ -11,10 +11,13 @@
...
 */
 
-$router->get('import/{importType}', 'ImportController@show')->name('import.show');
-$router->post('import/{importType}', 'ImportController@store')->name('import.store');
 Route::resource('failed-jobs', 'FailedJobController');
 
+$router->get('import/{importType}', 'ImportFileController@create')->name('import.show');
+$router->post('import/{importType}', 'ImportFileController@store')->name('import.store');
+
+Route::resource('import-files', 'ImportFileController');
```
mendelB: Would we want to limit this to the `index` route with an `only` parameter?

app/Http/Controllers/ImportFileController.php
aaronschachter: Eventually we'll show an import file, e.g. `/import-files/123`. This page is the upload webform, so renamed it to `create` for consistency with the Rogue web [ActionsController](https://github.com/DoSomething/rogue/blob/78cfb1abcdd590afd253fb4e9b8d9e83831bbb1f/app/Http/Controllers/Web/ActionsController.php#L37).

PR #127, opened 2020-03-06
Adds console command to download and import a Rock The Vote report

aaronschachter:
### What's this PR do?

This pull request continues on RTV API integration (#126) by adding a `rtv:import {$reportId}` that downloads and imports the Rock The Vote report for a given ID, if its status is `complete`.

It also adds an `options` text field to the `import_files` table, to save and display any parameters passed when importing a file.

Example console output:

```

vagrant@homestead:~/Code/chompy$ php artisan rtv:import 1966
[2020-03-06 19:44:42] local.INFO: Checking status of report 1966  

[2020-03-06 19:44:42] local.INFO: Report 1966 status is complete  

[2020-03-06 19:44:45] local.INFO: Downloaded report 1966  

[2020-03-06 19:44:45] local.INFO: STARTING rock-the-vote IMPORT {"options":"Array
(
    [report_id] => 1966
)
"} 

[2020-03-06 19:44:45] local.INFO: Broadcasting [Chompy\Events\LogProgress] on channels [importer] with payload:
{
    "message": "Total rows to chomp: 1774",
    "type": "general",
    "progressValue": null
}  

```
Example import files:

Screen Shot 2020-03-06 at 11 54 26 AM

### How should this be reviewed?

👀 

### Any background context you want to provide?

In a future pull request(s), we'll make requests to the `POST /reports` endpoint in the RTV API to create a report and pass its ID to this command.

### Relevant tickets

References [Pivotal #171656723](https://www.pivotaltracker.com/story/show/171656723).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Added appropriate feature/unit tests.

PR #128, opened 2020-03-07
Adds support to create Rock The Vote reports via API

aaronschachter:
### What's this PR do?

This pull request continues to climb the automated RTV import mountain, picking up from #126, #127 and adding a new `app/Services/RockTheVote` function, `createReport`,  that executes a [`POST /reports` API request](https://rock-the-vote.github.io/Voter-Registration-Tool-API-Docs/#reports). It also adds:

* A `rock_the_vote_reports` table, for keeping track of reports we've created and whether we need to import them (or already did)

* A web form at `/rock-the-vote/imports/create` to test the function by passing `since` and `before` parameters, which is what we'll need eventually for generating reports hourly

Screen Shot 2020-03-07 at 12 36 30 PM

Can test creating a report via the webform, it won't affect much except adding another report to the list of reports in our Rock The Vote portal (Tej said this is a non-issue). Testing will be a bit trickier once we want to start importing these reports once created, but that's for another PR.

Next up will be starting on importing the report once it's status is `complete` (the statuses a report has before it's done are `queued`, `building`, and `merging`), making use of the `imported_at` field to determine which reports we need to import when completed.

### Relevant tickets

References [Pivotal #171656723](https://www.pivotaltracker.com/story/show/171656723).

### Checklist

### DIFF
--- a/app/Models/RockTheVoteReport.php
+++ b/app/Models/RockTheVoteReport.php
@@ -35,4 +35,27 @@ class RockTheVoteReport extends Model
     public static $indexes = [
         'status',
     ];
+
+    /**
+     * Logs a Rock The Vote Reported created via API request.
+     *
+     * @param array $response
+     * @param string $since
+     * @param string $before
+     * @return RockTheVoteReport
+     */
+    public static function createFromApiResponse($response, $since = null, $before = null)
+    {
+        // Parse response to find the new Rock The Vote Report ID.
+        $statusUrlParts = explode('/', $response->status_url);
+        $reportId = $statusUrlParts[count($statusUrlParts) - 1];
+
+        // Log our created report in the database, to keep track of reports requested.
+        return static::create([
+            'id' => $reportId,
+            'since' => $since,
+            'before' => $before,
+            'status' => $response->status,
+        ]);
+    }
 }
--- a/app/Models/RockTheVoteReport.php
+++ b/app/Models/RockTheVoteReport.php
@@ -21,7 +21,7 @@ class RockTheVoteReport extends Model
         'status',
         'since',
         'before',
-        'imported',
+        'imported_at',
     ];
 
     /**
--- a/app/Models/RockTheVoteReport.php
+++ b/app/Models/RockTheVoteReport.php
@@ -37,7 +37,7 @@ class RockTheVoteReport extends Model
     ];
 
     /**
-     * Logs a Rock The Vote Reported created via API request.
+     * Logs a Rock The Vote Report that we've created via API request.
      *
      * @param array $response
      * @param string $since
--- a/app/Models/RockTheVoteReport.php
+++ b/app/Models/RockTheVoteReport.php
@@ -30,7 +30,6 @@ class RockTheVoteReport extends Model
      * @var array
      */
     protected $dates = [
-        'status',
         'dispatched_at',
     ];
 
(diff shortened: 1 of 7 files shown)


resources/views/pages/rock-the-vote-reports/show.blade.php
```diff
@@ -12,6 +12,11 @@
     <p>
         Total rows: <strong>{{$report->record_count}}</strong>
     </p>
+    @if ($report->status === 'building')
+    <p>
+        Progress: <strong>{{round(($report->current_index * 100) / $report->record_count)}}% </strong>(processed <strong>{{$report->current_index}}</strong> rows)
```
DFurnes: This could be nice to pull out into a method or computed property on the model!

```php
public function getPercentageAttribute()
{
    return round(($this->current_index * 100) / $this->record_count);
}
```

```blade
Progress: <strong>{{$report->percentage}}% </strong>(processed <strong>{{$report->current_index}}</strong> rows)
```

app/Models/RockTheVoteReport.php
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var array
+     */
+    protected $fillable = [
+        'id',
+        'status',
+        'since',
+        'before',
```
DFurnes: Since these are dates, do we want to add them to the [`$dates`](https://laravel.com/docs/7.x/eloquent-mutators#date-mutators) property on the model?

database/migrations/2020_03_06_000200_create_rock_the_vote_reports_table.php
```diff
@@ -0,0 +1,36 @@
...
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('rock_the_vote_reports', function (Blueprint $table) {
+            $table->integer('id');
+            $table->string('status')->index();
+            $table->string('since')->nullable();
+            $table->string('before')->nullable();
```
DFurnes: Do we want to store these as `dateTime` columns?

database/migrations/2020_03_06_000200_create_rock_the_vote_reports_table.php
```diff
@@ -0,0 +1,36 @@
...
+            $table->timestamps();
+            $table->date('imported_at')->nullable();
```
aaronschachter: I'm considering potentially renaming this field to `dispatched_at` -- it aims to save the time when we add a job to the queue to begin the import process (and wouldn't necessarily mean we imported the report if the download failed, for example)

resources/views/pages/rock-the-vote-reports/show.blade.php
aaronschachter: As is, this data is from a Rock the Vote API request to the `/GET registrant_reports/:id` endpoint, and is not displaying the info from the model. I wanted to save that change for a future PR to help keep this reviewable, but I think it'd be a sweet change to make once we do start displaying model info here.

app/Models/RockTheVoteReport.php
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * This array is manually maintained. It does not necessarily mean that
+     * any of these are actual indexes on the database... but they should be!
+     *
+     * @var array
+     */
+    public static $indexes = [
+        'status',
+        'dispatched_at',
+    ];
```
aaronschachter: I'm not so sure this is correct, or necessary. In the migration, I added a compound index on both `status` and `dispatched_at` to find any reports that are complete but not yet dispatched.

Also started a discussion in [Slack](https://dosomething.slack.com/archives/CUQMTP0RM/p1583779493003800?thread_ts=1583776494.002800&cid=CUQMTP0RM) about whether to rely on a job to keep checking the status of a report (vs a scheduled task that queries our DB)

resources/views/pages/rock-the-vote-reports/show.blade.php
DFurnes: Oh right!

PR #131, opened 2020-03-12
Import RTV report after creating 

aaronschachter:
### What's this PR do?

This pull request refactors the import RTV console command (`app/Console/Commands/ImportRockTheVoteReport`) introduced in #127 as a queued job (moving it to (`app/Jobs/ImportRockTheVoteReport`), and dispatches the job after creating a report via web (#128). 

It also adds some extra meta fields to our `RockTheVoteReport` model to store the progress of report generation locally as the import job checks the RTV API for the report status, and modifies the `show` view to display the model instead of making an API request. Refs https://github.com/DoSomething/chompy/pull/128#discussion_r389750611

Screen Shot 2020-03-12 at 7 20 05 AM

### How should this be reviewed?

👀 

### Any background context you want to provide?

This could really use some test coverage! I've been testing locally by creating real reports which is not ideal. I'm going to add coverage in a new PR because it requires adding `browser-kit-testing` (see https://github.com/DoSomething/chompy/pull/130#issuecomment-597918994) and looking to keep this PR easier to review. 

### Relevant tickets

References [Pivotal #171656723](https://www.pivotaltracker.com/story/show/171656723).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Added appropriate feature/unit tests.

PR #132, opened 2020-03-13
Adds tests for RTV report creation

aaronschachter:
### What's this PR do?

This pull request branches off of #131 to add test coverage for creating and importing a new Rock The Vote report via the form at `/rock-the-vote/reports/create`. Refs #126, #127, #128 

👀 

Not sure why we never got the User factory working properly and use `forceCreate` everywhere, but for now I'm following by example in the codebase and keeping it in -- saving refactoring it for another PR to help keep this (very big) PR on the hopefully reviewable side.

### Relevant tickets

References [Pivotal #171656723](https://www.pivotaltracker.com/story/show/171656723).

### Checklist


app/Jobs/ImportRockTheVoteReport.php
```diff
@@ -72,4 +72,17 @@ public function handle()
...
+    /**
+     * Returns the parameters passed to this job.
+     *
+     * @return array
+     */
+    public function getParameters()
+    {
+        return [
+            'report' => $this->report,
+            'user' => $this->user,
```
DFurnes: Should we return `user_id` here rather than the full User model?

tests/Http/Web/RockTheVoteReportTest.php
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Test creating and importing a Rock The Vote Report via web.
+     *
+     * @return void
+     */
+    public function testCreateRockTheReportFormSubmission()
+    {
+        Bus::fake();
+
+        $this->rockTheVoteMock->shouldReceive('createReport')->andReturn((object) [
```
DFurnes: Personally, I find it a little cleaner to use [`json_decode`](https://www.php.net/manual/en/function.json-decode.php)'s `$assoc` flag to return these as an associative array (and I think in general that's the pattern we follow elsewhere). That way you don't need to remember to cast this as an object elsewhere!

tests/Http/Web/RockTheVoteReportTest.php
```diff
@@ -0,0 +1,45 @@
...
+
+            $this->assertEquals($params['report']->id, 17);
+            $this->assertEquals($params['user'], $admin);
+
+            return true;
+        });
+
+        // Verify redirect to new Rock the Vote report.
+        $this->assertEquals(302, $response->status());
+        $this->assertTrue(Str::contains($response->getTargetUrl(), '/rock-the-vote/reports/17'));
```
DFurnes: I think we could use Laravel's [`assertRedirect`](https://laravel.com/docs/5.7/http-tests#assert-redirect) to simplify this a bit.

tests/Jobs/ImportRockTheVoteReportTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $this->rockTheVoteMock->shouldReceive('getReportStatusById')->andReturn((object) [
+            'status'=> 'building',
+            'record_count' => 117,
+            'current_index' => 3,
+        ]);
+        $this->rockTheVoteMock->shouldNotReceive('getReportByUrl');
+
+        $importRockTheVoteReportJob = new ImportRockTheVoteReport($user, $report);
+
+        $importRockTheVoteReportJob->handle();
```
DFurnes: It might be simpler to use the `::dispatch` syntax to do this in one shot:

```
ImportRockTheVoteReport::dispatch($user, $report);
```

I think manually instantiating & calling `handle` might also bite us if we ever use dependency injection in this job.

tests/Jobs/ImportRockTheVoteReportTest.php
```diff
@@ -0,0 +1,96 @@
...
+
+            $this->assertEquals($report, $params['report']);
+            $this->assertEquals($user, $params['user']);
+
+            return true;
+        });
+
+        Bus::assertNotDispatched(ImportFileRecords::class, function () {
+            return true;
+        });
```
DFurnes: Since you're just returning true here, this should be safe to omit:

```php
Bus::assertNotDispatched(ImportFileRecords::class);
```

tests/Jobs/ImportRockTheVoteReportTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $this->assertEquals($report->current_index, 3);
+        $this->assertEquals($report->user_id, $user->northstar_id);
+
+        Bus::assertDispatched(ImportRockTheVoteReport::class, function ($job) use (&$report, &$user) {
+            $params = $job->getParameters();
+
+            $this->assertEquals($report, $params['report']);
+            $this->assertEquals($user, $params['user']);
+
+            return true;
```
DFurnes: This might make more sense as the "success criteria" for the [bus assertion](https://laravel.com/docs/5.7/mocking#bus-fake):

```php
Bus::assertDispatched(ImportRockTheVoteReport::class, function ($job) use (&$report, &$user) {
    $params = $job->getParameters();
    
    return $params['report'] == $report && $params['user'] == $user;
});

```

tests/Unit/Model/RockTheVoteReportTest.php
```diff
@@ -0,0 +1,64 @@
...
+        $report = RockTheVoteReport::createFromApiResponse($apiResponse, $since, $before);
+
+        $this->assertEquals($report->id, 17);
+        $this->assertEquals($report->since, $since);
+        $this->assertEquals($report->before, $before);
+        $this->assertEquals($report->status, 'queued');
+        $this->assertEquals($report->user_id, null);
+        $this->assertDatabaseHas('rock_the_vote_reports', [
+            'id' => 17,
+        ]);
```
DFurnes: This seems redundant since you're testing the same thing against the Eloquent model directly above.

tests/Jobs/ImportRockTheVoteReportTest.php
aaronschachter: Could it already be biting us? I had tried calling the job this way, following by example in our [CallPower test](https://github.com/DoSomething/chompy/blob/674ad32ebffcc0bec959f1d63e4fb69fa3bb9579/tests/Http/CallPowerTest.php#L92), but because we're calling `Bus::fake` in this test -- the assertions for verifying the report model update fail:
```
        ImportRockTheVoteReport::dispatch($user, $report);

        $this->assertEquals($report->status, 'building');
```

fails with
```
1) Tests\Http\Web\ImportRockTheVoteReportTest::testReportStatusNotComplete
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'queued'
+'building'
```

app/Jobs/ImportRockTheVoteReport.php
aaronschachter: I went with a full User model because that's what we're expecting when we [dispatch a `ImportFileRecords` job](https://github.com/DoSomething/chompy/blob/674ad32ebffcc0bec959f1d63e4fb69fa3bb9579/app/Jobs/ImportFileRecords.php#L51) if the report status is complete.

tests/Http/Web/RockTheVoteReportTest.php
aaronschachter: Do you mean making this change over in  the `app/Services/RockTheVote:: createReport ` function, to return an array instead of an object?

PR #136, opened 2020-03-16
Refactors RTV report creation

aaronschachter:
### What's this PR do?

This pull request renames the `createFromApiResponse` method to `createViaApi`, changing the function to execute the `POST /reports` request to RTV API, instead of expecting an API response as a parameter. 

We'll need to create a report from a console command to run our import hourly, so this PR makes the change to DRY executing the API request.

👀 

Opening this now before adding the code for the scheduled import job to help keep the PR's reviewable.

### Relevant tickets

### Checklist

### DIFF
--- a/app/Models/RockTheVoteReport.php
+++ b/app/Models/RockTheVoteReport.php
@@ -2,6 +2,7 @@
 
 namespace Chompy\Models;
 
+use Chompy\Services\RockTheVote;
 use Illuminate\Database\Eloquent\Model;
 
 class RockTheVoteReport extends Model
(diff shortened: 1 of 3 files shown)


app/Models/RockTheVoteReport.php
```diff
@@ -52,15 +53,19 @@ class RockTheVoteReport extends Model
...
-     * Logs a Rock The Vote Report that we've created via API request.
+     * Creates a Rock The Vote Report via API request and saves to storage.
      *
-     * @param object $response
      * @param datetime $since
      * @param datetime $before
      * @return RockTheVoteReport
      */
-    public static function createFromApiResponse($response, $since, $before)
+    public static function createViaApi($since = null, $before = null)
```
katiecrane: We won't always be using `$since` and `$before`?

aaronschachter: For our purposes... we do always pass these parameters currently, but the API doesn't require them (if we wanted to generate a report for all VR's for example). This felt cleaner than passing `null` as parameter values in the tests (hence this change)

katiecrane: Oh nice, sounds good!

PR #137, opened 2020-03-16
Adds Console command to import RTV data from the past hour

aaronschachter:
### What's this PR do?

This pull request supports automating the Rock the Vote import entirely, adding the following features/tests:

* A `CreateRockTheVoteReport` job, that accepts a time range as parameters and creates a new RTV report and dispatches for import upon success.

* An `import:rock-the-vote` console command that dispatches the `CreateRockTheVoteReport` job, passing the last hour as the time range to create a report for. We'll use the Heroku Scheduler to run this command hourly.

👀 

This PR borrows the [`mockTime` function from Rogue ](https://github.com/DoSomething/rogue/blob/b42cd7183075017cf7ac7b9ab8b0daad160905ed/tests/WithMocks.php#L70) to test the parameters passed in the `ImportRockTheVoteCommandTest`.

### Relevant tickets

### Checklist


tests/Jobs/RockTheVote/CreateRockTheVoteReportTest.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        Bus::fake();
+
+        $this->rockTheVoteMock->shouldReceive('createReport')->andThrow(new Exception(500));
+
+        $createRockTheVoteReportJob = new CreateRockTheVoteReport('2020-03-08 00:00:00', '2020-03-08 00:01:00');
+
+        $this->expectException(Exception::class);
+
+        $createRockTheVoteReportJob->handle();
```
aaronschachter: I tried to use `createRockTheVoteReportJob::dispatch` here, but the error wasn't thrown. I suspect this is similar to the thread over in https://github.com/DoSomething/chompy/pull/132#discussion_r392334282, where it seems calling `Bus::fake` prevents us from testing the job via dispatch cc @DFurnes.

app/Jobs/RockTheVote/CreateRockTheVoteReport.php
```diff
@@ -0,0 +1,76 @@
...
+     *
+     * @return array
+     */
+    public function handle()
+    {
+        info('Creating report', ['before' => $this->before, 'since' => $this->since]);
+
+        $this->report = RockTheVoteReport::createViaApi($this->since, $this->before);
+
+        ImportRockTheVoteReport::dispatch(null, $this->report);
```
Vmack: Everything looks fine! Just a question about line 60...why is null included on the dispatch?

aaronschachter: We're sending `null` for the `user` parameter because this job is being dispatched by the console command. We set it to the authenticated user when we create a report via the web over in the `RockTheVoteReportController`:
https://github.com/DoSomething/chompy/blob/2c8fa724eda9e8077b1c8f5c69e91a12869aa622/app/Http/Controllers/Web/RockTheVoteReportController.php#L50

Thank you for reviewing! 😸

Vmack: ohhh ok got it! And you're welcome @aaronschachter 🙂

PR #139, opened 2020-03-18
Increase overlapping interval for hourly reports

aaronschachter:
### What's this PR do?

Now that the second automated import has run in production, I manually created a report between the last row of the 1st automated report  2020-03-18 13:28:36 and the last row of the 2nd report : 2020-03-18 13:43:07-- to see if we missed any records. This report generated 159 rows, which is very weird. There's also a disconnect between the dates I'm passing in the API and the dates returned in the report.

This PR increases the overlap time to 30 minutes just to be on the safe side, to make sure we don't miss any rows.

👀 

⏳ 

### Relevant tickets

### Checklist

### DIFF
--- a/app/Console/Commands/ImportRockTheVoteCommand.php
+++ b/app/Console/Commands/ImportRockTheVoteCommand.php
@@ -33,8 +33,8 @@ class ImportRockTheVoteCommand extends Command
         $before = Carbon::now();
         $since = clone $before;
 
-        // ... and after an hour ago (adding two overlapping minutes to ensure we don't miss any within the internal).
-        $since->subHours(1)->subMinutes(2);
+        // ... and after an hour ago (adding 30 overlapping minutes to ensure we don't miss any within the internal).
+        $since->subHours(1)->subMinutes(30);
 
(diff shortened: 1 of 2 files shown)


mendelB:
Is it at all possible that RTV has some delay on processing/including the rows in the report per DateTime, so that we might not be getting _all_ the latest data when generating the report? (Though I would assume that wouldn't been an issue before the automation 🤔)

aaronschachter:
The last machine generated report before I deployed this change (ID 2219) had its final row with a date of  020-03-18 14:36:37 -0400

The next machine generated report (ID 2222) after deploying this change, had its first row with a date of 2020-03-18 14:16:59 -0400 -- so it looks like this worked as intended. 

What's not ideal is that the dates within the rows don't match the range of the `since` and `before` parameters we are passing. Report 2222 was created with Since: `2020-03-18 18:10:18`  and Before: `2020-03-18 19:40:18` -- seems like it's a timezone discrepency? Their [API docs](https://rock-the-vote.github.io/Voter-Registration-Tool-API-Docs/#reports) suggest passing a UTC datetime format - which I thought we're doing currently? cc @DFurnes

PR #140, opened 2020-03-23
Adds is_valid_mobile helper

aaronschachter:
### What's this PR do?

This pull request adds a very scrappy first pass at a `is_valid_mobile` helper, checking for the value `000-000-0000`.  We received a registration with `000-000-0000` set as the mobile, which otherwise looks otherwise valid -- this should cause the failed job to succeed the next time we retry it.

Screen Shot 2020-03-23 at 4 15 40 PM

👀 

Can add more to this helper if it comes up, for now this is the only mobile number validation issue I've seen.

### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -35,3 +35,18 @@ function is_anonymous_mobile($mobile)
         '+8656696',
     ]);
 }
+
+/**
+ * Determines if a mobile number is valid.
+ *
+ * @TODO: Add phone number validation lib, or DRY this with Northstar validation through Gateway.
+ * @see https://git.io/Jvy0A
+ * For now, RTV is passing through a 000-000-0000 number that causes user creation to fail.
+ *
+ * @param string $mobile
+ * @return bool
+ */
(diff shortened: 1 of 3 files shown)


app/helpers.php
```diff
@@ -35,3 +35,19 @@ function is_anonymous_mobile($mobile)
         '+8656696',
     ]);
 }
+
+/**
+ * Determines if a mobile number is valid.
+ *
+ * @TODO: Add phone number validation lib, or DRY this with Northstar validation through Gateway.
+ * @see https://git.io/Jvy0A
```
mendelB: Agreed - this would def be neat!

mendelB:
@aaronschachter Am I understanding correctly that we're just addressing an edge case bypassing RTV's validation?

aaronschachter:
> Am I understanding correctly that we're just addressing an edge case bypassing RTV's validation?

That's exactly what's this is for - thanks for taking a look!

PR #147, opened 2020-03-30
Adds SMS topic for new users + integration tests for updating existing users

aaronschachter:
### What's this PR do?

This pull request adds a `sms_subscription_topics` array value to the `RockTheVoteRecord->userData` payload if the user has opted in to SMS messaging, populating it for new users that are created via import. (refs https://github.com/DoSomething/northstar/pull/998)

It also adds a TODO for updating existing users with this field, and adds integration for the existing update logic before I start on modifying the sms subscriptions of existing users in a separate pull request - to help keep this reviewable.

👀 

You'll see in the commits that I was going to add a function called `updatePostIfExists`, similar to `updateUserIfExists` -- but didn't like that I'd need to pass the `$rogue` class instance into the function to use it. Does anyone know the history of why we need the `Rogue` service at all, vs calling Gateway like we do for the Northstar calls? I couldn't figure out how the `$rogue` instance gets passed as an argument the job's `handle` function. cc @katiecrane @DFurnes 

### Relevant tickets

References [Pivotal #171848031](https://www.pivotaltracker.com/story/show/171848031).

### Checklist

### DIFF
--- a/app/RockTheVoteRecord.php
+++ b/app/RockTheVoteRecord.php
@@ -40,7 +40,10 @@ class RockTheVoteRecord
 
         if ($this->userData['mobile']) {
             // Note: Not a typo, this column name does not have the trailing question mark.
-            $this->userData['sms_status'] = str_to_boolean($record['Opt-in to Partner SMS/robocall']) ? 'active' : 'stop';
+            $smsOptIn = str_to_boolean($record['Opt-in to Partner SMS/robocall']);
+
+            $this->userData['sms_status'] = $smsOptIn ? 'active' : 'stop';
+            $this->userData['sms_subscription_topics'] = $smsOptIn ? explode(',', $config['user']['sms_subscription_topics']) : [];
         }
 
         $this->postData = [
--- a/tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
+++ b/tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
@@ -118,7 +118,7 @@ class ImportRockTheVoteRecordTest extends TestCase
         $this->mockGetNorthstarUser([
             'voter_registration_status' => 'registration_complete',
         ]);
-        $this->rogueMock->shouldNotReceive('createUser');
+        $this->northstarMock->shouldNotReceive('createUser');
         $this->northstarMock->shouldNotReceive('updateUser');
         $this->rogueMock->shouldReceive('getPost')->andReturn([
             'data' => [
--- a/tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
+++ b/tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
@@ -120,4 +120,36 @@ class ImportRockTheVoteRecordTest extends TestCase
 
         $job->updateUserIfChanged($user);
     }
+
+    /**
+     * Test that status should update when update value has higher priority than the current.
+     *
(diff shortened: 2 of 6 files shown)


app/RockTheVoteRecord.php
```diff
@@ -40,7 +40,10 @@ public function __construct($record, $config = null)
 
         if ($this->userData['mobile']) {
             // Note: Not a typo, this column name does not have the trailing question mark.
-            $this->userData['sms_status'] = str_to_boolean($record['Opt-in to Partner SMS/robocall']) ? 'active' : 'stop';
+            $smsOptIn = str_to_boolean($record['Opt-in to Partner SMS/robocall']);
+
+            $this->userData['sms_status'] = $smsOptIn ? 'active' : 'stop';
```
mendelB: Sorry, this is irrelevant to this change but just curious if there's any special reason we explicitly set a `stop` status, instead of just _not_ setting 'em to `active`?

tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
```diff
@@ -70,4 +70,102 @@ public function testDoesNotCreateOrUpdatePostIfCompletedPostFound()
...
+    {
+        $statusHierarchy = [
+            'uncertain',
+            'ineligible',
+            'unregistered',
+            'confirmed',
+            'register-OVR',
+            'register-form',
+            'registration_complete',
+        ];
```
mendelB: Do you think this potentially indicates the (non-critical) need for this hierarchy to be a config or some formalized declaration somewhere?

app/RockTheVoteRecord.php
aaronschachter: The reason is that we are saving their mobile number, regardless if they opt-in or not. So by not opting and setting them to `stop`, we can be sure we won't message them. I've been following the existing import logic.  There could be an option to change this and simply not save their mobile number if they didn't opt in, but the decision to save their mobile even if they don't subscribe was likely made in order to try to assign one Northstar ID per person (if I subscribe later from my phone and the RTV never captured it, I'm now in the age-old 2 Northstar users predicament)

tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
aaronschachter: I like the idea moving this into the config! Can revisit in a future PR.

app/RockTheVoteRecord.php
mendelB: Ah that makes sense re the number! I assumed that we _only_ message folks who are actively `active`, and the `stop` status is a way to indicate an 'active' unsubscribe but an empty status would _also_ not get messaging.

aaronschachter: I'd think an empty status wouldn't get messaging too -- but since users won't be checking the opt-in box, setting to `stop` makes sense (vs null).

mendelB: Ah gotcha - so in a sense, there _is_ an active 'unsubscribe' or 'non-subscribe' happening here that we're faithfully assigning.

aaronschachter:
> Does anyone know the history of why we need the Rogue service at all, vs calling Gateway like we do for the Northstar calls?

Just kidding, there is no Rogue client in [Gateway](https://github.com/DoSomething/gateway/tree/master/src).

PR #148, opened 2020-03-31
Avoid double importing Rock the Vote records

aaronschachter:
### What's this PR do?

This pull request modifies the RTV import to only process a record if it hasn't been logged in the `rock_the_vote_logs` table. 

The code changes are pretty small, this PR mostly adds test coverage to ensure these changes work as expected. What's changed:

* Create the `rock_the_vote_logs` entry after the `ImportRockTheVoteRecord` job has finished all of its tasks (either creating user or getting existing user and checking for updates + creating post or getting existing post and checking for updates)

* Adds a `RockTheVoteLog::getByRecord` function to search for whether a row has been imported already

    * Adds a compound index to the `rock_the_vote_logs` table to optimize search

* Extracts creating and updating posts into smaller `createPost` and `updatePostIfChanged` functions to improve readability of the `handle` function

* Adds a `RockTheVoteRecord` helper, `getPostDetails` to easily find row details saved into a `post.details` string property 

It hasn't been a big deal to re-process a record thus far, because we only update a user or post if a given record has a `Status` column value with higher priority than the user's `voter_registration_status` field or their voter registration post's `status` field.

However, once we modify this job to update the SMS preferences for users we import -- it will be important to NOT re-import a row, as the user might unsubscribe from SMS by the time we re-import the row. Our automated import may end up fetching a row a second time, because we purposely look for rows 30 minutes before the last job finished -- in order to avoid potentially missing rows in between jobs (see #139). 

### Relevant tickets

Prep work for subscribing existing users in [Pivotal #171848124](https://www.pivotaltracker.com/story/show/171848124).

### Checklist

### DIFF
--- a/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
+++ b/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
@@ -62,26 +62,14 @@ class ImportRockTheVoteRecord implements ShouldQueue
         ]);
 
         if (! $existingPosts['data']) {
-            $post = $rogue->createPost(array_merge(['northstar_id' => $user->id], $this->postData));
+            $this->createPost($user);
+        } else {
+            $post = $existingPosts['data'][0];
 
-            info('Created post', ['post' => $post['data']['id'], 'user' => $user->id]);
+            info('Found post', ['post' => $post['id'], 'user' => $user->id]);
 
-            return;
+            $this->updatePostIfChanged($post);
         }
-
-        $post = $existingPosts['data'][0];
-
-        info('Found post', ['post' => $post['id'], 'user' => $user->id]);
-
-        if (! self::shouldUpdateStatus($post['status'], $this->postData['status'])) {
-            info('No changes to update for post', ['post' => $post['id']]);
-
-            return;
-        }
-
-        $rogue->updatePost($post['id'], ['status' => $this->postData['status']]);
-
-        info('Updated post', ['post' => $post['id'], 'status' => $this->postData['status']]);
 
         // @TODO: Update log with imported_at.
     }
@@ -146,16 +134,27 @@ class ImportRockTheVoteRecord implements ShouldQueue
 
         info('Created user', ['user' => $user->id]);
 
-        $post = app(Rogue::class)->createPost(array_merge([
-            'northstar_id' => $user->id,
-        ], $this->postData));
-
-        info('Created post', ['post' => $post['data']['id'], 'user' => $user->id]);
+        $this->createPost($user);
 
         // @TODO: Pass imported_at
         RockTheVoteLog::createFromRecord($this->record, $user, $this->importFileId);
 
         $this->sendUserPasswordResetIfSubscribed($user);
+    }
+
+    /**
+     * Creates new post with given Northstar user and record post data.
+     *
+     * @param NorthstarUser
+     * @return void
+     */
+    private function createPost($user)
+    {
+        $post = app(Rogue::class)->createPost(array_merge([
+            'northstar_id' => $user->id,
+        ], $this->postData));
+
+        info('Created post', ['post' => $post['data']['id'], 'user' => $user->id]);
     }
 
     /**
@@ -209,6 +208,23 @@ class ImportRockTheVoteRecord implements ShouldQueue
         gateway('northstar')->asClient()->updateUser($user->id, $payload);
 
         info('Updated user', ['user' => $user->id, 'voter_registration_status' => $this->userData['voter_registration_status']]);
+    }
+
+    /**
+     * @param array $post
+     * @return void
+     */
+    private function updatePostIfChanged($post)
+    {
+        if (! self::shouldUpdateStatus($post['status'], $this->postData['status'])) {
+            info('No changes to update for post', ['post' => $post['id']]);
+
+            return;
+        }
+
+        app(Rogue::class)->updatePost($post['id'], ['status' => $this->postData['status']]);
+
+        info('Updated post', ['post' => $post['id'], 'status' => $this->postData['status']]);
     }
 
     /**
--- a/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
+++ b/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
@@ -52,7 +52,11 @@ class ImportRockTheVoteRecord implements ShouldQueue
             return $this->handleNewUser();
         }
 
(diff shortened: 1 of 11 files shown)


app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
```diff
@@ -134,20 +130,40 @@ private function getUser($id, $email, $mobile)
...
     /**
-     * Creates new user with record user data.
+     * Creates new user and post with record data.
      *
-     * @param array $data
-     * @return NorthstarUser
+     * @return void
      */
-    private function createUser()
+    private function handleNewUser()
```
weerd, line 137: Why the change in method name? Prior name seemed to follow with `createPost` and using the typical CRUD verbs.

aaronschachter: Because this function creates a new user, but also a new post, a log record, and sends the password reset email if user subscribed. I thought it followed with along with `handle` function nicely, but could potentially rename to something like `createUserAndPost`

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
```diff
@@ -203,6 +219,23 @@ private function updateUserIfChanged($user)
...
+        if (! self::shouldUpdateStatus($post['status'], $this->postData['status'])) {
+            info('No changes to update for post', ['post' => $post['id']]);
+
+            return;
+        }
+
+        app(Rogue::class)->updatePost($post['id'], ['status' => $this->postData['status']]);
+
+        info('Updated post', ['post' => $post['id'], 'status' => $this->postData['status']]);
+    }
```
weerd, line 237: I think we could keep this method named `updatePost()` and since currently used in only one place move the logic to check if it should update back where line 81 is (`$this->updatePostIfChanged($post);`). Then this method can solely deal w/ updating a post.

Alternatively, having to simpler methods, `updatePostIfChanged()` which does the check and if needs to update then calls `updatePost()`, so each method has a specific responsibility. Allows the `updatePost()` to be called without needing to do the check if that's every a thing.

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
```diff
@@ -203,6 +219,23 @@ private function updateUserIfChanged($user)
         info('Updated user', ['user' => $user->id, 'voter_registration_status' => $this->userData['voter_registration_status']]);
     }
 
+    /**
+     * @param array $post
+     * @return void
+     */
```
weerd, line 225: Missing DocBlock description.

database/factories/ModelFactory.php
```diff
@@ -1,11 +1,24 @@
...
 use Faker\Generator;
+use Chompy\Models\RockTheVoteLog;
 use Chompy\Models\RockTheVoteReport;
 
+$factory->define(RockTheVoteLog::class, function (Generator $faker) {
+    return [
+        'finish_with_state' => 'No',
+        'import_file_id' => $this->faker->randomDigitNotNull,
+        'pre_registered' => 'No',
+        'started_registration' => '2020-02-22 19:16:32 -0500',
```
weerd, line 12: Would it be better to use `Carbon` for this faked date? You can set it to do dates in past if needed, but seems better than maybe introducing a typo into this very specific date. 🤔

app/RockTheVoteRecord.php
```diff
@@ -163,21 +163,23 @@ private function parseVoterRegistrationStatus($rtvStatus, $rtvFinishWithState)
...
-
-        $importantKeys = [
-            'Tracking Source',
-            'Started registration',
-            'Finish with State',
-            'Status',
-            'Pre-Registered',
-            'Home zip code',
-        ];
+        $config = ImportType::getConfig(ImportType::$rockTheVote);
```
weerd, line 166: I've seen this in a few spots now and it seems really confusing. Upon further inspection it looks like it's just grabbing the config for Rock The Vote. Why can't we just use the ole `config()` helper which is pretty straightforward? This series of static properties and methods just seems confusing to me. This can essentially be just:

```php
$config = config('import.rock_the_vote');
```

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
aaronschachter: I prefer keeping the check if it should update in here, so we don't have a nested if/else from where it was removed. Also keeps things consistent with the `updateUserIfChanged`, which will soon get more complex once we need to look for changes to SMS status and potentially add an existing user's mobile number.

As far as having two different methods `updatePostIfChanged` and `updatePost` -- having `updatePost` be called without needing to do the check won't ever be a thing because we only update it from this one job.  We could split out the methods at that point if we'd ever need to.

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
weerd: So to me that highlights that maybe the location that calls this method should do the other things instead of mixing it all into one method. It's a better practice to keep methods like this doing a single thing.

Instead in the `handle()`method above, something like this?

```php

if (! $user) {

  $user = $this->createUser();

  $this->createPost($user);

  RockTheVoteLog::createFromRecord($this->record, $user, $this->importFileId);

  $this->sendUserPasswordResetIfSubscribed($user);
}
```

aaronschachter: I think that makes sense - will make this change.

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
weerd: Gotcha, ok.

> which will soon get more complex

However this is what I want to help mitigate. The more complex these functions get the harder they are to test (or require more complex tests), which is why I'm always more in favor of smaller single use functions as much as possible.

aaronschachter: I definitely hear you -- I'm just weary of all if/else logic happening within the `handle. I'll definitely keep this in mind once it's time to work on updating the SMS preferences. Thanks for reviewing so thoroughly!

weerd: Ya, defs. Lots of if/else logic is not ideal either; there are ways around it but likely require a little refactoring! :)

PR #151, opened 2020-04-02
Adds contains_phone column to RTV logs

aaronschachter:
### What's this PR do?

This pull request adds a boolean `contains_phone` field to `rock_the_vote_logs`, tracking whether or not a RTV row contains a value for the `Phone` field. This way, we can avoid potentially re-subscribing someone if they happen to unsubscribe during the voter registration process.

Example:

* Our hourly RTV import receives a row for Alice with status `Step 4` that has opted her in to SMS (phone field is filled out, as well as opting in to DS texts). We process this record, and send Alice a welcome text 📲 

* Alice receives our welcome text, unsubscribes by texting back STOP 🚫, and then continues to complete their voter registration

* In the next hourly import, we receive a new RTV row for Alice for status `Complete`, which will still contain the same phone field and opt-in field values per above. We need the ability to check for any previous updates for this registration that would have subscribed her, in order to avoid re-subscribing. 

👀 

The check for whether a user registration has already updated someone's SMS preferences will be to query `rock_the_vote_logs` for the user's `user_id`,  the `started_registration` datetime, and a true value for `contains_phone`. If a `rock_the_vote_logs` entry is found, we should not update the user's SMS subscription -- otherwise we're in the clear to update their Northstar profile (and will be saving a `rock_the_vote_log` with `contains_phone` to true, so we won't make any other changes during the current registration). 

I hope this makes sense, but please 💬 me should you have any ❓ 

### Relevant tickets

### Checklist

### DIFF
--- a/app/Models/RockTheVoteLog.php
+++ b/app/Models/RockTheVoteLog.php
@@ -14,6 +14,7 @@ class RockTheVoteLog extends Model
      * @var array
      */
     protected $fillable = [
+        'contains_phone',
         'finish_with_state',
         'import_file_id',
         'pre_registered',
@@ -44,6 +45,7 @@ class RockTheVoteLog extends Model
         $info = $record->getPostDetails();
 
         $rockTheVoteLog = self::create([
+            'contains_phone' => isset($record->userData['mobile']),
             'import_file_id' => $importFile->id,
             'finish_with_state' => $info['Finish with State'],
             'pre_registered' => $info['Pre-Registered'],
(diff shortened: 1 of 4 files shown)


katiecrane:
This is mostly adding this column right, the actual work to check if we should change the SMS subscription status hasn't happened right? Making sure I'm not missing something!

aaronschachter:
That's right! Wanted to just get the column in to help keep things reviewable once it's time to start updating users' SMS subscriptions.

katiecrane:
@aaronschachter a true hero! perfect!

PR #154, opened 2020-04-08
Create voter-reg post per unique "Started registration"

aaronschachter:
### What's this PR do?

This pull request modifies the Rock the Vote `getPost` function to check for a matching "Started registration" value in the post details when returning a result. As a result (ha ha), we'll create a new `voter-reg` post every time a new "Started registration" is present for a user in a RTV report, instead of upserting a single post for all "Started registration" values.

👀 

This change corresponds to an org-wide change to base our topline metric of total voter registrations on gross registrations, where every completed voter registration application counts as a “voter registration” -- instead of unique users completing voter registration applications. See the [Voting Registration Problems/Solutions doc](https://docs.google.com/document/d/1An-JD0Tp6bQIvqYHfDpZvTjNwPkBxB8soB22ZLJ-lG8/edit#bookmark=id.lt5xqhgftxns) for more context.

### Relevant tickets

References [Pivotal #171737617](https://www.pivotaltracker.com/story/show/171737617)

### Checklist

- [x] Documentation added for new features/changed endpoints.

### DIFF
--- a/database/faker/FakerRockTheVoteReportRow.php
+++ b/database/faker/FakerRockTheVoteReportRow.php
@@ -34,7 +34,9 @@ class FakerRockTheVoteReportRow extends Base
     }
 
     /**
+     * Return a mock Started registration field value.
      *
+     * @return string
      */
     public function rockTheVoteStartedRegistration()
     {
--- a/database/faker/FakerRockTheVoteReportRow.php
+++ b/database/faker/FakerRockTheVoteReportRow.php
@@ -25,11 +25,19 @@ class FakerRockTheVoteReportRow extends Base
             'Phone' => $this->generator->phoneNumber,
             'Finish with State' => 'Yes',
             'Pre-Registered' => 'No',
-            'Started registration' => Carbon::now()->format('Y-m-d H:i:s O'),
+            'Started registration' => $this->rockTheVoteStartedRegistration(),
             'Status' => 'Step 1',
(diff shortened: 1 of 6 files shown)


database/faker/FakerRockTheVoteReportRow.php
```diff
@@ -25,11 +25,21 @@ public function rockTheVoteReportRow($data = [])
...
     }
+
+    /**
+     * Return a mock Started registration field value.
+     *
+     * @return string
+     */
+    public function rockTheVoteStartedRegistration($numDaysToSubtract = 0)
+    {
+        return Carbon::now()->subDays($numDaysToSubtract)->format('Y-m-d H:i:s O');
```
DFurnes, line 43: Is the main reason for this generator the non-standard date-format? Wondering if `daysAgoInRockTheVoteFormat()` might read a little clearer? If only they used normal ISO-8601!

aaronschachter: Yep, that's the only reason for the non-standard and I don't mind renaming. It wouldn't really break any of our tests to mock with normal ISO-8601, but seemed better to try to mock the data to match as best as possible.

aaronschachter:
> This is a monster PR!

Indeed, thanks for reviewing! The majority of the changes were tests, been wondering about whether there are cleaner ways to cover all scenarios.. but for another refactor I think.

PR #155, opened 2020-04-09
Starts on updating SMS information for existing users

aaronschachter:
### What's this PR do?

This pull request adds a feature flag to begin working on support for updating an existing user's SMS information, starting with the following changes:

* Adds a `RockTheVoteLog::hasAlreadyUpdatedSmsSubscription` function that checks to see whether a given registration record has already been imported with a phone number (refs #151)

* If `hasAlreadyUpdatedSmsSubscription` returns true, don't update the user's SMS information when importing the current registration record. We would have already updated the SMS information for the user in the previous registration record that contained a phone number, we don't want to opt them in again if they happened to unsubscribe since the last import ran that subscribed them.

* If `hasAlreadyUpdatedSmsSubscription` returns false, only write the user's `mobile` field when they don't have one set. Refs [Pivotal thread](https://www.pivotaltracker.com/story/show/171848124/comments/212652843).

👀 

We won't want to enable this feature flag until:
1.  We add support for updating a user's SMS and status subscription topics based on whether the SMS opt-in is checked
2. Rock The Vote adds the SMS opt-in checkbox to their registration form

Will take on the work of the first item in another PR in hopes of keeping this one reviewable.

### Relevant tickets

### Checklist

### DIFF
--- a/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
+++ b/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
@@ -262,7 +262,7 @@ class ImportRockTheVoteRecord implements ShouldQueue
          * If we have a log for this registration that contains phone, we've already
          * updated user's subscription for this registration and should not proceed.
          */
-        if (RockTheVoteLog::hasSubmittedPhone($this->record, $user)) {
+        if (RockTheVoteLog::hasAlreadyUpdatedSmsSubscription($this->record, $user)) {
             return [];
         }
 
--- a/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
+++ b/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
@@ -224,22 +224,22 @@ class ImportRockTheVoteRecord implements ShouldQueue
     }
 
     /**
-     * Get fields and values to update given user with.
+     * Update Northstar user with record data.
      *
-     * @return array
+     * @return void
      */
     public function updateUserIfChanged(NorthstarUser $user)
     {
         $payload = [];
 
         if (self::shouldUpdateStatus($user->voter_registration_status, $this->userData['voter_registration_status'])) {
-            $payload = array_only($this->userData, ['voter_registration_status']);
+            $payload['voter_registration_status'] = $this->userData['voter_registration_status'];
         }
 
         if (config('import.rock_the_vote.update_user_sms_enabled') == 'true') {
             $payload = array_merge($payload, $this->getUserSmsSubscriptionUpdatePayload($user));
         }
-        
+
         if (! count($payload)) {
             info('No changes to update for user', ['user' => $user->id]);
 
--- a/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
+++ b/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
@@ -259,10 +259,13 @@ class ImportRockTheVoteRecord implements ShouldQueue
         }
 
         /**
-         * @TODO: If the RTV status is Complete: check whether we have a log for Step 3/Step 4.
-         * It would mean we've already processed this user's subscription preferences, and
-         * shouldn't update them again.
+         * If we have a log for this registration that contains phone, we've already
+         * updated user's subscription for this registration and should not proceed.
          */
+        if (RockTheVoteLog::hasSubmittedPhone($this->record, $user)) {
+            return [];
+        }
+
         $result = array_only($this->userData, ['sms_status', 'sms_subscription_topics']);
 
         // Save mobile if we don't have it currently stored on the user.
--- a/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
+++ b/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
@@ -233,9 +233,9 @@ class ImportRockTheVoteRecord implements ShouldQueue
     /**
      * Get fields and values to update given user with.
      *
-     * @param NorthstarUser $user
+     * @return array
      */
-    private function getUpdateUserPayload($user)
+    public function getUpdateUserPayload(NorthstarUser $user)
     {
         $payload = [];
 
@@ -253,10 +253,9 @@ class ImportRockTheVoteRecord implements ShouldQueue
     /**
      * Get fields and values to update given user with if their SMS subscription has changed.
      *
-     * @param NorthstarUser $user
      * @return array
      */
-    public function getUpdateUserSmsSubscriptionPayload($user)
+    public function getUpdateUserSmsSubscriptionPayload(NorthstarUser $user)
     {
         $result = [];
 
@@ -265,10 +264,7 @@ class ImportRockTheVoteRecord implements ShouldQueue
             return $result;
         }
 
-        /**
-         * If we have a log for this registration that contains phone, we've already updated the
-         * user's subscription for this registration, and should not proceed.
-         */
+        // We don't need to update user's SMS subscription if we already did for this registration.
         if (RockTheVoteLog::hasAlreadyUpdatedSmsSubscription($this->record, $user)) {
             return $result;
         }
--- a/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
+++ b/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
@@ -76,7 +76,14 @@ class ImportRockTheVoteRecord implements ShouldQueue
             return;
         }
 
-        $this->updateUserIfChanged($user);
+        $updateUserPayload = $this->getUpdateUserPayload($user);
+
+        if (count($updateUserPayload)) {
+            gateway('northstar')->asClient()->updateUser($user->id, $updateUserPayload);
+            info('Updated user', ['user' => $user->id]);
+        } else {
+            info('No changes to update for user', ['user' => $user->id]);
+        }
 
         if ($post = $this->getPost($user)) {
             $this->updatePostIfChanged($post);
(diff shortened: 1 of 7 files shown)


config/import.php
```diff
@@ -68,6 +70,9 @@
             'register-form',
             'registration_complete',
         ],
+        // Whether to update SMS profile information for an existing user.
+        'update_user_sms_enabled' => env('ROCK_THE_VOTE_UPDATE_USER_SMS_ENABLED', 'false'),
+        // Constants to use when creating a new user.
```
weerd, line 75: Helpful comments! 👌

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
```diff
@@ -76,7 +76,14 @@ public function handle(Rogue $rogue)
             return;
         }
 
-        $this->updateUserIfChanged($user);
+        $updateUserPayload = $this->getUpdateUserPayload($user);
```
weerd, line 79: I was struggling a bit with following some of the code in this PR, and I think it mostly has to do with the naming 😬 `$updateUserPayload` sounds like it's a boolean; deciding whether it should or shouldn't update the payload.

Maybe rename the variable to `$updatedUserData`?

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
```diff
@@ -224,29 +231,52 @@ public static function shouldUpdateStatus($currentStatus, $newStatus)
...
 
     /**
-     * Update Northstar user with record data.
+     * Get fields and values to update given user with.
      *
-     * @param NorthstarUser $user
+     * @return array
      */
-    private function updateUserIfChanged($user)
+    public function getUpdateUserPayload(NorthstarUser $user)
```
weerd, line 238: Same here, especially with what the method used to be named was a bit confusing to me. Would maybe `getUpdatedUserData()` be clearer?

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
aaronschachter: I'm not too keen on calling it `updated` when it hasn't been updated yet, this is the payload we're going to use for making a request to update the user. I don't agree that `updateUserPayload` sounds like a boolean -- I thought the `Payload` would indicate it's an object (or an associate array in PHP land). Maybe something like `userFieldValuesToChange`?

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
weerd: I think that when reading it as a method name, it's not common to read it as "get updateUser() payload". Where you have another method's name in this method's name itself. It ends up sounding odd and confusing. I think with most of our method names, we aim to try and have them read as a sentence so you don't have the added need to figure out if it's referring to another method in the name itself, depending on how you read it.

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
weerd: Ah, ok. See part of the confusion with knowing if it's updated yet is because the naming sounded odd to me, along with `getUpdateUserPayload()`. Which is why I was stuck on it trying to decipher what was going on.

Ya, payload does indicate an object/array but "update user payload?" sounds like a question to me, needing to decide whether the payload should or shouldn't be updated; because it's not obvious to read it as "updateUser() payload" as in the payload for the `updateUser()` method.

weerd: The previous method of `updateUserIfChanged()` was more straightforward to me, because it indicates that if the user info changed, lets go ahead and update it. Is there a reason we can't keep that method as it was named and then rework the logic in there to deal with the feature flag etc?

aaronschachter: That's fair enough.  Would just switching the order to `userUpdatePayload` work? Then the other function could be named `getUserUpdatePayload`

weerd: @aaronschachter not sure if you saw my comment about `updateUserIfChanged()` since I think we submitted comments at the exact same time!

Muscle High Five

DFurnes: Just popping in from #notify-applications… 🤓 I can definitely see how this method could be read two entirely different ways. It's like optical illusions where it can look like something's spinning clockwise or counter-clockwise just depending on how it "clicks" in your brain.

Renaming this to `userUpdatePayload` and `getUserUpdatePayload` definitely feels like an improvement.

I also like @weerd's idea of keeping these in one method to sidestep the naming concerns altogether (since as far as I know this method isn't used anywhere else, right?). That might be a little easier to read, you can step through all the logic that goes into updating a user (or not) sequentially.

aaronschachter: The discussion over in https://github.com/DoSomething/chompy/pull/148#discussion_r401756658 was one prompt that led me to refactor `updateUserIfChanged` into smaller function, to test each one. The part 2 of this PR would be to change a user's `sms_status` and `sms_subscription_topics` per row values, which adds even more permutations to test. For those, I was thinking of another smaller function like `getUserSmsStatusAndSubscriptionTopicsPayload` to test those variations vs all of them in one method.

It does make testing each smaller function easier, which is why these functions are public, so we can just test each payload-getter. For sure, it could be refactored to test all as `updateUserIfChanged` though... I can try that first to see if there were any gotchas with going that route, vs renaming multiple times/functions.

weerd: Could you maybe do a bit of both... keep `updateUserIfChanged()` since it makes sense and then in there it does all the calls to the smaller functions. Those smaller functions could still be tested independently, and the `updateUserIfChanged()` is just a convenience method that kind of collects all that logic into an easier to call place. Does that make sense? Maybe this is already what you mean tho 😅

aaronschachter: Ah, I thought it was cleaner to break out of each function when we have smaller functions that return payloads to construct -- e.g. There's no mobile here? Ok return an empty array -- vs a bunch of if/else -- also per https://github.com/DoSomething/chompy/pull/148#discussion_r401793957 review.

aaronschachter: > `updateUserIfChanged()` is just a convenience method that kind of collects all that logic into an easier to call place.

That's what the `getUserUpdatePayload` is doing -- collecting all the logic. All `updateUserIfChanged` would look like is
```
if ($data = getUserUpdatePayload($user)) {
   gateway->northstar->updateUser($data)
}
```
which is where it seemed straight forward enough to have this logic in the `handle` as is in this PR vs a separate `updateUserIfChanged`. Per my last comment, being able to exit out of each payload building function was what felt cleaner than nested if/else statements and calling Northstar update user function at the end.

weerd: I'm thinking something like this:

```php
public function updateUserIfChanged($user)
{
  $payload = [];

  $payload = array_merge($payload, methodThatReturnsArrayIfMobileChanged());

  $payload = array_merge($payload, methodThatReturnsArrayIfStatusChanged());

  $payload = array_merge($payload, methodThatReturnsArrayIfSomeOtherFieldChanged());

  if(count($payload)) {
    // send request to update the user's information
  } else {
    // no update needed
  }
}
```

Pseudo-ish code, but hopefully that makes sense. Each separate method can still be tested separately. This also still alleviates the amount of logic in just the `updateUserIfChanged()` method, but spread out into more specific, smaller functions.

aaronschachter: Makes sense! Made this change, thanks for the example.

tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
```diff
@@ -166,6 +171,65 @@ public function testUpdatesUserIfShouldChangeStatus()
         ]);
     }
 
+    /**
+     * Test that user mobile is not updated when row has a phone, user does not have a mobile, and
+     * update_user_sms_enabled config is false.
+     *
+     * @return void
+     */
+    public function testUserUpdatePayloadDoesNotContainsMobileIfUpdateUserSmsConfigIsDisabled()
```
weerd, line 180: 👀 typo: "testUserUpdatePayloadDoesNotContain~~s~~MobileIfUpdateUserSmsConfigIsDisabled"

tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
```diff
@@ -166,6 +171,65 @@ public function testUpdatesUserIfShouldChangeStatus()
...
+    {
+        \Config::set('import.rock_the_vote.update_user_sms_enabled', 'true');
+
+        $user = new NorthstarUser([
+            'id' => $this->faker->northstar_id,
+            'voter_registration_status' => 'step-1',
+        ]);
+        $phoneNumber = $this->faker->phoneNumber;
+        $row = $this->faker->rockTheVoteReportRow([
+            'Opt-in to Partner SMS/robocall' => 'Yes',
```
weerd, line 217: that's one heck of a row field name, haha

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
```diff
@@ -226,17 +226,19 @@ public static function shouldUpdateStatus($currentStatus, $newStatus)
     /**
      * Update Northstar user with record data.
      *
-     * @param NorthstarUser $user
+     * @return void
      */
-    private function updateUserIfChanged($user)
+    public function updateUserIfChanged(NorthstarUser $user)
```
weerd, line 231: As a future refactor, I still think we could get rid of the conditionals and and move that logic into respective called methods, but not a big deal for now.

PR #156, opened 2020-04-14
Finishes updating SMS information for existing users in RTV import

aaronschachter:
### What's this PR do?

This pull request modifies the Rock The Vote import to update an existing user's `sms_status` and `sms_subscription_topics` based on whether they opt-in to receive voting related texts from DS when they register to vote (Note: this opt-in isn't visible on the RTV site yet, which is why these changes that were started in #155 are still feature flagged)

The expected behavior for changes to the  `sms_status` and `sms_subscription_topics` have been [added to the documentation in this PR](https://github.com/DoSomething/chompy/blob/rtv-sms-status/docs/imports/rock-the-vote.md#sms-status), copy/pasting here:

**SMS Status**

* If an existing user opts-in to voting-related SMS messaging from DS via RTV form, we update sms_status as active if current value is either null, less, stop, or undeliverable.

* If an existing user opts-out of voting-related SMS messaging from DS via RTV form, we update sms_status as stop if current value is null or undeliverable.

**SMS Subscription Topics**

* If an existing user opts-in to voting-related SMS messaging from DS via RTV form, we add the voting topic to their sms_subscription_topics if it doesn't exist.

* If an existing user opts-out of voting-related SMS messaging from DS via RTV form, we remove the voting topic from their sms_subscription_topics if it exists

The bulk of this PR are unit tests to make sure the update values are expected per a user's current `sms_status` and `sms_subscription_topics` fields. It's a little tedious to test each SMS status value and opting in and out, but also felt like the 100% way to feel all bases were covered. Trying to put together a test CSV to manually spot check this against your local Northstar users db collection feels way more painful and error-prone.

Besides needing the RTV checkbox visible before we enable the feature flag, we also want to backfill the `sms_subscription_topics` of all existing SMS subscribers with  `general` and `voting` -- so we can honor an existing user's preference to opt-out ONLY of voting texts, and not all general texts from DS.

### Relevant tickets

References [Pivotal #171848124](https://www.pivotaltracker.com/story/show/171848124), [Slack thread](https://dosomething.slack.com/archives/C03T8SDMS/p1586798992063300).

### Checklist

### DIFF
--- a/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
+++ b/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
@@ -319,15 +319,13 @@ class ImportRockTheVoteRecord implements ShouldQueue
         }
 
         /**
-         * If user hasn't opted-in and has current subscription topics, remove if present.
-         *
-         * Note: we can get away with just checking for the config import topics because there we
-         * only have one.
+         * If user hasn't opted-in and has current subscription topics, remove any import topics
+         * from user's current subscription topics.
          */
         $result[$key] = [];
 
         foreach ($currentSmsTopics as $topic) {
-            if ($topic != config('import.rock_the_vote.user.sms_subscription_topics')) {
+            if (! in_array($topic, explode(',', config('import.rock_the_vote.user.sms_subscription_topics')))) {
                 array_push($result[$key], $topic);
             }
         }
--- a/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
+++ b/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
@@ -260,19 +260,17 @@ class ImportRockTheVoteRecord implements ShouldQueue
      */
     public function getUserSmsSubscriptionUpdatePayload(NorthstarUser $user)
     {
-        $payload = [];
-
         // If registration does not have a mobile provided, there is nothing to update.
         if (! $this->userData['mobile']) {
-            return $payload;
+            return [];
         }
 
         // We don't need to update user's SMS subscription if we already did for this registration.
         if (RockTheVoteLog::hasAlreadyUpdatedSmsSubscription($this->record, $user)) {
-            return $payload;
+            return [];
         }
 
-        $payload = array_merge($payload, $this->parseMobileChangeForUser($user));
+        $payload = $this->parseMobileChangeForUser($user);
         $payload = array_merge($payload, $this->parseSmsSubscriptionTopicsChangeForUser($user));
         $payload = array_merge($payload, $this->parseSmsStatusChangeForUser($user));
 
--- a/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
+++ b/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
@@ -271,15 +271,27 @@ class ImportRockTheVoteRecord implements ShouldQueue
             return $payload;
         }
 
+        $payload = array_merge($payload, $this->parseMobileChangeForUser($user));
         $payload = array_merge($payload, $this->parseSmsSubscriptionTopicsChangeForUser($user));
         $payload = array_merge($payload, $this->parseSmsStatusChangeForUser($user));
 
-        // Update user's mobile only if we currently don't have one saved.
-        if (! $user->mobile) {
-            $payload['mobile'] = $this->userData['mobile'];
+        return $payload;
+    }
+
+    /**
+     * Returns payload to update mobile if user does not currently have one saved.
+     *
+     * @return array
+     */
+    public function parseMobileChangeForUser(NorthstarUser $user)
+    {
+        $key = 'mobile';
+
+        if ($user->{$key}) {
+            return [];
         }
 
-        return $payload;
+        return [$key => $this->userData[$key]];
     }
 
     /**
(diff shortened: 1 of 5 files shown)


app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
```diff
@@ -258,26 +260,101 @@ public function updateUserIfChanged(NorthstarUser $user)
...
+     * @return array
+     */
+    public function parseSmsSubscriptionTopicsChangeForUser(NorthstarUser $user)
+    {
+        $fieldName = 'sms_subscription_topics';
+        $currentSmsTopics = ! empty($user->{$fieldName}) ? $user->{$fieldName} : [];
+
+        // If user opted in to SMS, add the import topics to current topics.
+        if ($this->smsOptIn) {
+            return [$fieldName => array_unique(array_merge($currentSmsTopics, $this->userData[$fieldName]))];
```
DFurnes, line 308: Should this merge `$currentSmsTopics` with `config('import.rock_the_vote.user.sms_subscription_topics')` instead? What is in `$this->userData[$fieldName]` at this point?

DFurnes: Also – this doesn't need to block merging, but it'd be nice to handle this in Northstar with a `SmsTopicController` (like [we do for email subscription topics](https://github.com/DoSomething/northstar/blob/master/app/Http/Controllers/SubscriptionUpdateController.php)) – then we don't need to worry about merging anything in this application (or race-conditions if another change to topics comes in at the same time).

So this could be something like:

```php
public function updateSmsSubscriptionTopicsForUser(NorthstarUser $user)
{
    if ($this->smsOptIn) {
        $this->addSmsTopics($user, config('import.rock_the_vote.user.sms_subscription_topics'));
    } else {
        $this->removeSmsTopics($user, config('import.rock_the_vote.user.sms_subscription_topics'));
    }
}

```

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
```diff
@@ -258,26 +260,101 @@ public function updateUserIfChanged(NorthstarUser $user)
...
+        if ($this->smsOptIn && in_array($currentSmsStatus, [
+            SmsStatus::$less,
+            SmsStatus::$pending,
+            SmsStatus::$stop,
+        ])) {
+            return [$fieldName => $importSmsStatus];
         }
 
-        return $result;
+        return [];
```
DFurnes, line 357: Is this something we could handle in Northstar – if we add a topic for someone and they're set to `STOP`, we could automatically set their status to `active`? (I believe  that's how we handle changes to email topics, right?)

app/SmsStatus.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Chompy;
+
+class SmsStatus
+{
+    /**
+     * @var string
+     */
+    public static $active = 'active';
```
DFurnes, line 10: These should probably be [class constants](https://www.php.net/manual/en/language.oop5.constants.php).

tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
```diff
@@ -494,6 +499,298 @@ public function testSmsSubscriptionPayloadIsEmptyIfUserAlreadyHasMobile()
...
+
+        $this->assertEquals([], $result);
+    }
+
+    /**
+     * Tests that import topic is added from SMS topics if user opts-in to RTV SMS.
+     *
+     * @return void
+     */
+    public function testParseSubscriptionTopicsChangeIfUserHasTopicsAndOptsOut()
```
DFurnes, line 769: Should this be `AndOptsIn`?

tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
```diff
@@ -494,6 +499,298 @@ public function testSmsSubscriptionPayloadIsEmptyIfUserAlreadyHasMobile()
...
+     * parseSmsSubscriptionTopicsChangeForUser
+     * ---------------------------------------
+     */
+
+    /**
+     * Tests that import topic is removed from SMS topics if user opts-out of RTV SMS.
+     *
+     * @return void
+     */
+    public function testParseSubscriptionTopicsChangeIfUserOptsIn()
```
DFurnes, line 721: Should this be `IfUserOptsOut`?

tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
```diff
@@ -3,12 +3,14 @@
...
 
 use Tests\TestCase;
+use Chompy\SmsStatus;
 use Chompy\RockTheVoteRecord;
 use Chompy\Models\ImportFile;
 use Chompy\Models\RockTheVoteLog;
 use Chompy\Jobs\ImportRockTheVoteRecord;
 use DoSomething\Gateway\Resources\NorthstarUser;
 
+// @TODO: Replace all hardcoded RTV field names with RockTheVoteRecord constants to DRY.
```
DFurnes, line 13: I generally prefer _not_ to have my tests rely on application code (like these constants), since it means an unintentional change in a constant might not fail a test. And this is probably more subjective, but I actually prefer to have my tests be less DRY – since it means you can read them more like independent "stories", rather than having to jump around to different files to piece together helpers and constants.

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
aaronschachter: > Should this merge $currentSmsTopics with config('import.rock_the_vote.user.sms_subscription_topics') instead? What is in $this->userData[$fieldName] at this point

`$this->userData['sms_subscription_topics]` contains the same `config('import.rock_the_vote.user.sms_subscription_topics')`. It's convenient to store it in the userData for creating a user, where we can just pass the whole array to `northstar->createUser($this->userData)`. For a user that doesn't opt-in, the `$this->userData[sms_subscription_topics]` is an empty array -- which is why it's reading from the config when removing topics.

I did think about the email subscriptions endpoint, but it seemed a bit overkill since we're updating a few different fields for a user per record, and it's cleaner to send a single update request (vs different API requests to modify each field, e.g. subscription, mobile, status, voter registration)

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
aaronschachter: This touches on https://github.com/DoSomething/northstar/pull/1008 -- subscribing to SMS has the variation that you could be re-subscribing with a `less` or `active` status. The email subscription topic is easier to automate since the email subscription status is boolean.

app/SmsStatus.php
aaronschachter: This was following the example of our `ImportType` class, introduced per https://github.com/DoSomething/chompy/pull/59#discussion_r258555699. I'm happy to refactor both classes in a future PR

DFurnes: Oh oops!! I guess I wasn't thinking as clearly back then. 🙈

PR #157, opened 2020-04-17
Adds form to test RTV import

aaronschachter:
### What's this PR do?

This pull request adds support to test the Rock The Vote import for a single record via webform if `source=test` query parameter is passed to the RTV upload form (e.g. `/import/rock-the-vote?source=test`).

Screen Shot 2020-04-17 at 12 35 22 PM

### How should this be reviewed?

👀 

### Any background context you want to provide?

This makes it easier to significantly manually test the import, instead of modifying a CSV locally and uploading.  And it's also surfaced a bug already, from my local testing:

* There's a bug in #156 where a `GET /users/:id` API response isn't returning the `sms_subscription_topics` array in the payload, so it's never removing the `voting` subscription topic (this is another good example of when only testing via mocks can bite us, like in https://github.com/DoSomething/chompy/pull/153). This feature isn't turned on yet, so not a biggie

* ~~There's a bug where we are no longer saving birthdate in user creation. This should get fixed ASAP 🤕~~ -- Not a bug, see [Slack](https://dosomething.slack.com/archives/CTVPG6L4R/p1587153485493600?thread_ts=1587153236.493300&cid=CTVPG6L4R)

These bugs to be fixed in a future PR, to help keeps this reviewable. 

### Relevant tickets

References [Pivotal #171848124](https://www.pivotaltracker.com/story/show/171848124).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added appropriate feature/unit tests.

PR #158, opened 2020-04-17
RTV import cleanup

aaronschachter:
### What's this PR do?

This pull request addresses two items flagged in #157:

* Removes debugging a `$user` object, as the mystery was solved in https://github.com/DoSomething/northstar/pull/1010

* Clarifies which user fields are and are not imported via import (I thought we mistakenly stopped importing birthdate during the many surgeries this import has undergone)

👀 

🪑 

### Relevant tickets

References [Pivotal #171848124](https://www.pivotaltracker.com/n/projects/2417735/stories/171848124).

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/ImportFileController.php
+++ b/app/Http/Controllers/Web/ImportFileController.php
@@ -150,7 +150,7 @@ class ImportFileController extends Controller
         if ($importType === ImportType::$rockTheVote) {
             $row = [
                 'Email address' => $request->input('email'),
-                'Finish with State' => $request->input('finish_with_state'),
+                'Finish with State' => $request->input('finish_with_state') ?: 'No',
                 'First name' => $request->input('first_name'),
                 'Home address' => $request->input('addr_street1'),
                 'Home city' => $request->input('addr_city'),
--- a/app/Http/Controllers/Web/ImportFileController.php
+++ b/app/Http/Controllers/Web/ImportFileController.php
@@ -56,8 +56,8 @@ class ImportFileController extends Controller
                 'email' => $user->email,
                 'first_name' => $user->first_name,
                 'last_name' => $user->last_name,
-                'mobile' => $user->mobile,
-                'referral' => 'ads',
+                'phone' => $user->mobile,
+                'tracking_source' => 'source:test,source_details:ChompyUI',
                 'started_registration' => Carbon::now()->format('Y-m-d H:i:s O'),
             ];
         }
@@ -150,7 +150,7 @@ class ImportFileController extends Controller
         if ($importType === ImportType::$rockTheVote) {
             $row = [
                 'Email address' => $request->input('email'),
-                'Finish with State' => $request->input('status'),
+                'Finish with State' => $request->input('finish_with_state'),
                 'First name' => $request->input('first_name'),
                 'Home address' => $request->input('addr_street1'),
                 'Home city' => $request->input('addr_city'),
@@ -159,11 +159,11 @@ class ImportFileController extends Controller
                 'Last name' => $request->input('last_name'),
                 'Opt-in to Partner SMS/robocall' => $request->input('sms_opt_in') ?: 'No',
                 'Opt-in to Partner email?' => $request->input('email_opt_in') ?: 'No',
-                'Phone' => $request->input('mobile'),
+                'Phone' => $request->input('phone'),
                 'Pre-Registered' => $request->input('pre_registered') ?: 'No',
                 'Started registration' => $request->input('started_registration'),
                 'Status' => $request->input('status'),
-                'Tracking Source' => $request->input('referral'),
+                'Tracking Source' => $request->input('tracking_source'),
             ];
 
             $importFile = new ImportFile();
--- a/app/Http/Controllers/Web/ImportFileController.php
+++ b/app/Http/Controllers/Web/ImportFileController.php
@@ -57,7 +57,7 @@ class ImportFileController extends Controller
                 'first_name' => $user->first_name,
                 'last_name' => $user->last_name,
                 'mobile' => $user->mobile,
-                'referral' => 'user:'.$userId.',source:test,source_detail:ChompyUI',
+                'referral' => 'ads',
                 'started_registration' => Carbon::now()->format('Y-m-d H:i:s O'),
             ];
         }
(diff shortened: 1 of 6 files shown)


docs/imports/rock-the-vote.md
```diff
@@ -102,27 +102,49 @@ If an existing user has a null `mobile` profile field and provides a phone numbe
...
+
+- Email Subscription
+
+  - If user opts-in to email messaging from DS, user is subscribed with `community` topic and an Activate Account email is sent.
+
+- SMS Subscription
+
+  - If user opts-in to SMS messging from DS, user is subscribed to the `voting` SMS topic with status `active`.
+
+  - If user opts-out of SMS messging from DS, user's `voting` SMS topic is removed if exists. SMS status will be set to `active` if was previously set to `stop` or `undeliverable`.
```
katiecrane, line 139: This is in the new user section, so I don't think this bullet point needs to account for anything already existing or being set as something else.

app/Http/Controllers/Web/ImportFileController.php
```diff
@@ -57,7 +57,7 @@ public function create($importType)
                 'first_name' => $user->first_name,
                 'last_name' => $user->last_name,
                 'mobile' => $user->mobile,
-                'referral' => 'user:'.$userId.',source:test,source_detail:ChompyUI',
+                'referral' => 'ads',
```
katiecrane, line 60: Want to make sure I'm reading this right - this is for test imports?

docs/imports/rock-the-vote.md
aaronschachter: Thanks for catching this copy 🍝 error! Will update.

app/Http/Controllers/Web/ImportFileController.php
aaronschachter: It is -- I've been meaning to update this example to include `source` and `source_details` per above, but just remove the `user` query parameter -- which would have to be constantly cleared from the text input to test other users beyond that parameter value.

katiecrane: I think I'm still missing something - didn't you say you wanted to keep `source` and `source_detail`?

aaronschachter: Sorry! Forgot to explicitly mention that I want to update this example tracking source too (got caught up in something else)

katiecrane: My turn to apologize - what is the tracking source? is that the same as `referral` here?

aaronschachter: Updated the names of these inputs to correspond to the column names found in the RTV CSV. Totally valid question - we do use the tracking source to determine whether the voter registration was created from a referral page, but also used it for general `source` and `source_details` regardless of its a referral -- so went with `tracking_source`

app/Http/Controllers/Web/ImportFileController.php
```diff
@@ -159,11 +159,11 @@ public function store(Request $request, $importType)
                 'Last name' => $request->input('last_name'),
                 'Opt-in to Partner SMS/robocall' => $request->input('sms_opt_in') ?: 'No',
                 'Opt-in to Partner email?' => $request->input('email_opt_in') ?: 'No',
-                'Phone' => $request->input('mobile'),
+                'Phone' => $request->input('phone'),
```
mendelB, line 162: These and the other input field name updates are per updates on the RTV end I'm assuming?

docs/imports/rock-the-vote.md
```diff
@@ -102,27 +102,49 @@ If an existing user has a null `mobile` profile field and provides a phone numbe
...
+### SMS Subscription
 
-### SMS Subscription Topics
+If an existing user **opts-in** to voting-related SMS messaging from DS:
 
-If an existing user **opts-in** to voting-related SMS messaging from DS via RTV form, we add the `voting` topic to their `sms_subscription_topics` if it doesn't exist.
+- add the `voting` topic to `sms_subscription_topics` if it doesn't exist.
 
-If an existing user **opts-out** of voting-related SMS messaging from DS via RTV form, we remove the `voting` topic from their `sms_subscription_topics` if it exists.
+- update `sms_status` as `active` if current value is either null, `less`, `stop`, or `undeliverable`.
```
mendelB, line 115: This is very much out of this PR's scope (and my contextual knowledge) but this Q occurred to me while reading this. If a user has the 'general' `sms_subscription_topics`, and we update their status here from, say, `null`. Are we opting them into something more then what they intended in VR world?

docs/imports/rock-the-vote.md
```diff
@@ -102,27 +102,49 @@ If an existing user has a null `mobile` profile field and provides a phone numbe
...
+
+- Email Subscription
+
+  - If user opts-in to email messaging from DS, user is subscribed with `community` topic and an Activate Account email is sent.
+
+- SMS Subscription
+
+  - If user opts-in to SMS messging from DS, user is subscribed to the `voting` SMS topic with status `active`.
+
+  - If user opts-out of SMS messging from DS, user's `sms_status` will be set to `stop` and `sms_subscription_topics` will be set to empty.
```
mendelB, line 139: tiny typo -- `messging` -> `messaging`

app/Http/Controllers/Web/ImportFileController.php
aaronschachter: These renames are on our end -- I renamed these inputs on our test webform to better match the RTV field names, which are the array keys here.

mendelB: Ohhh lol right right! It's right there below in the PR 🤦

docs/imports/rock-the-vote.md
aaronschachter: We shouldn't ever have a scenario where a user has a `general` subscription topic but a null `sms_status` value. The plan is backfill all current SMS subscribers with both `general` and `voting` subscription topics. Anyone who will be opting-in via RTV who was previously unsubscribed will only receive the `voting` subscription topic. Does this answer your question?

mendelB: Ahh this does indeed, thank you!! (I guess we remove the topics if someone unsubscribes entirely?)

PR #159, opened 2020-04-21
Display user and post info after creating a test RTV import

aaronschachter:
### What's this PR do?

This pull request branches off of #158 to display the created/updated user and their voter registration post when submitting a test RTV import the webform added in #157

Screen Shot 2020-04-20 at 7 35 25 PM

:eyes:

It's helpful to manually run an import to verify Northstar and Rogue return the expected results. We mock these results in our API requests, but it's been valuable to test against the actual services to catch any unexpected behavior (e.g. https://github.com/DoSomething/chompy/pull/157#discussion_r410490921). Returning the created or updated user and post will make it easier to test (don't need to check Aurora/Rogue to verify expected results)

### Relevant tickets

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/ImportFileController.php
+++ b/app/Http/Controllers/Web/ImportFileController.php
@@ -150,7 +150,7 @@ class ImportFileController extends Controller
         if ($importType === ImportType::$rockTheVote) {
             $row = [
                 'Email address' => $request->input('email'),
-                'Finish with State' => $request->input('finish_with_state'),
+                'Finish with State' => $request->input('finish_with_state') ?: 'No',
                 'First name' => $request->input('first_name'),
                 'Home address' => $request->input('addr_street1'),
                 'Home city' => $request->input('addr_city'),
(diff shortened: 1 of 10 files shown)


app/Http/Controllers/Web/ImportFileController.php
```diff
@@ -174,13 +177,38 @@ public function store(Request $request, $importType)
             $importFile->import_type = $importType;
             $importFile->save();
 
+            $html = 'Created <a href="/import-files/'.$importFile->id.'">import #'.$importFile->id.'</a>.<ul>';
             $job = new ImportRockTheVoteRecord($row, $importFile);
+            $result = $job->handle();
```
DFurnes, line 182: I'd recommend using Laravel's [`dispatchNow`](https://laravel.com/docs/5.7/queues#synchronous-dispatching) to dispatch this (rather than manually invoking the `handle` method). That way, we don't break dependency injection if we end up relying on other classes in this job down the line.



```suggestion
            $result = ImportRockTheVoteRecord::dispatchNow($row, $importFile);
```

aaronschachter:
Thanks for reviewing! Made these changes, and held off on linking out to the different entities for a later PR should we want to add them.

Screen Shot 2020-04-21 at 8 23 13 PM

PR #162, opened 2020-04-24
Tests for updating an existing user's SMS subscription

aaronschachter:
### What's this PR do?

This pull request refactors the many tests for `parseSmsStatusChangeForUser` function to instead test for the `updateUserIfChanged` function -- which has different behavior based on whether the `UPDATE_USER_SMS_ENABLED` config variable is set. Now that we're looking to enable updating the SMS subscriptions for existing users via the RTV import, it's ideal to test that `updateUserIfChanged` does change the SMS subscription for existing users as we'd expect to when the feature flag is enabled. 

👀 

There are so many different tests because the import will behave differently based on an existing user's SMS status, and whether they opt-in.

The groundwork for all of these tests was laid out by the `parseSmsStatus` tests, since that has the most variation -- so it felt like it made sense to refactor these unit tests as more of a general integration test for updating the user, to hopefully keep this PR readable/reviewable.

These tests rely on mocking calls to the Northstar API -- there's also a related [separate document for a human to manually QA each scenario](https://docs.google.com/document/d/1wKg33xbLdrnNaJyNH0vmBaUN_eSFyqd8Ug4f-CGXCHc/edit?usp=sharing), should that help provide background context. I'll be finishing up that doc as I manually test each scenario myself via the Import Test form (#157) 

### Relevant tickets

### Checklist

### DIFF
--- a/tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
+++ b/tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
@@ -612,58 +612,84 @@ class ImportRockTheVoteRecordTest extends TestCase
     /**
      * @return void
      */
-    public function testUserSmsSubscriptionUpdateIfUserHasActiveSmsStatusAndOptsIn()
+    public function testEnabledUpdateUserSmsWhenUserHasActiveSmsStatusAndOptsIn()
     {
+        $this->enableUpdateUserSmsFeature(true);
+
         $mocks = $this->getMocksForParseSmsStatusChangeTest(SmsStatus::$active, true);
         $job = new ImportRockTheVoteRecord($mocks->row, factory(ImportFile::class)->create());
 
-        $result = $job->getUserSmsSubscriptionUpdatePayload($mocks->user);
+        $this->northstarMock->shouldNotReceive('updateUser');
 
-        $this->assertEquals([], $result);
+        $job->updateUserIfChanged($mocks->user);
+
+        $this->enableUpdateUserSmsFeature(false);
     }
 
     /**
      * @return void
      */
-    public function testUserSmsSubscriptionUpdateIfUserHasActiveSmsStatusAndOptsOut()
+    public function testEnabledUpdateUserSmsWhenUserHasActiveSmsStatusAndOptsOut()
     {
+        $this->enableUpdateUserSmsFeature(true);
+
         $mocks = $this->getMocksForParseSmsStatusChangeTest(SmsStatus::$active, false);
         $job = new ImportRockTheVoteRecord($mocks->row, factory(ImportFile::class)->create());
 
-        $result = $job->getUserSmsSubscriptionUpdatePayload($mocks->user);
+        $this->northstarMock->shouldReceive('updateUser')
+          ->with($mocks->user->id, [
+              // The 'voting' topic should be removed.
+              'sms_subscription_topics' => ['general'],
+          ])
+          ->andReturn($mocks->user);
 
-        $this->assertEquals([
-            // The 'voting' topic should be removed.
-            'sms_subscription_topics' => ['general'],
-        ], $result);
+        $job->updateUserIfChanged($mocks->user);
+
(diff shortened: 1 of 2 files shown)


tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
```diff
@@ -533,211 +533,284 @@ public function testSmsSubscriptionPayloadIsEmptyIfUserAlreadyHasMobile()
...
-     *
      * @return void
      */
-    public function testParseSmsStatusChangeIfUserHasNullSmsStatusAndOptsIn()
+    public function testEnabledUpdateUserSmsWhenUserHasNullSmsStatusAndOptsIn()
     {
-        $mocks = $this->getMocksForParseSmsStatusChangeTest(null, true);
+        $this->enableUpdateUserSmsFeature(true);
+
+        $mocks = $this->getMocksForUpdateUserSmsTest(null, true);
```
DFurnes, line 570: At first glance, I wasn't sure what `null, true` did here. I wonder if splitting this into two methods might be a bit more readable... maybe something like this? (It'd be even nicer if we could use model factories here, but I think that's probably a no-go since neither of these are actual Eloquent models...)

```php
$user = $this->getNorthstarUserMock(['sms_status' => null]);
$row = $this->getRockTheVoteRowMock($user, true);
```

tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
```diff
@@ -533,211 +533,284 @@ public function testSmsSubscriptionPayloadIsEmptyIfUserAlreadyHasMobile()
...
+          ->with($mocks->user->id, [
+              'sms_status' => SmsStatus::$active,
+              'sms_subscription_topics' => ['voting'],
+          ])
+          ->andReturn($mocks->user);
 
-        $this->assertEquals(['sms_status' => SmsStatus::$active], $result);
+        $job->updateUserIfChanged($mocks->user);
+
+        $this->enableUpdateUserSmsFeature(false);
```
DFurnes, line 582: Each test should be a fresh application environment, so it shouldn't be necessary to reset this to `false`. (If that _isn't_ the case & subsequent tests inherit the same application state, then we should fix that in `TestCase`!)

tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
```diff
@@ -533,211 +533,284 @@ public function testSmsSubscriptionPayloadIsEmptyIfUserAlreadyHasMobile()
...
+    public function testEnabledUpdateUserSmsWhenUserHasNullSmsStatusAndOptsIn()
     {
-        $mocks = $this->getMocksForParseSmsStatusChangeTest(null, true);
+        $this->enableUpdateUserSmsFeature(true);
+
+        $mocks = $this->getMocksForUpdateUserSmsTest(null, true);
         $job = new ImportRockTheVoteRecord($mocks->row, factory(ImportFile::class)->create());
 
-        $result = $job->parseSmsStatusChangeForUser($mocks->user);
+        $this->northstarMock->shouldReceive('updateUser')
```
DFurnes, line 573: This is subjective, but I think I'd move these lines up above the `$job = ` lines so all the mocks are together.

aaronschachter: Great suggestion! Much cleaner 🧹

tests/Jobs/RockTheVote/ImportRockTheVoteRecordTest.php
aaronschachter: Definitely felt like I was losing readability by using one function to create both mocks, but it was also so tedious to write these tests, that having this function felt like a real timesaver. I'd love to split these out into unit and integration a little more some time, i could see creating a child class that could handle creating the mock user/rows for these subscription updates on setup to potentially DRY.

PR #167, opened 2020-05-18
Check if errors method exists before sending to log

aaronschachter:
### What's this PR do?

This pull request fixes a bug introduced by #166 where sometimes we see:  `Uncaught Error: Call to undefined method Symfony\Component\Debug\Exception\FatalErrorException::errors()`

👀 

🩹 

### Relevant tickets

References [Pivotal #172805635](https://www.pivotaltracker.com/story/show/172805635).

### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -49,6 +49,7 @@ function is_anonymous_mobile($mobile)
  */
 function is_valid_mobile($mobile)
 {
+    return true;
     // This phone number has been passed before and fails Northstar validation.
     if ($mobile == '000-000-0000') {
         return false;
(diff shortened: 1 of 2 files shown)


app/helpers.php
```diff
@@ -49,6 +49,7 @@ function is_anonymous_mobile($mobile)
  */
 function is_valid_mobile($mobile)
 {
+    return true;
```
katiecrane, line 52: I think this line might be leftover from some testing you were doing maybe?

aaronschachter: Thank you for catching! Indeed it is/was.

PR #177, opened 2020-07-13
Upgrade to Laravel 6.0

DFurnes:
### What's this PR do?

This pull request upgrades Chompy to Laravel 6.0, the latest long-term service release.

1️⃣ Updated Laravel 5.7 to Laravel 5.8, following the [upgrade guide](https://laravel.com/docs/5.8/upgrade) in 41fd1c7.

2️⃣ Updated from Laravel 5.8 to 6.0, following the [upgrade guide](https://laravel.com/docs/6.0/upgrade) in a94b9e5. Because Laravel 6.0 requires Carbon 2, I also upgraded to [Gateway 3.0](https://github.com/DoSomething/gateway/releases/tag/v3.0.0). This involved updating [to the new `getUser` signature](https://github.com/DoSomething/gateway/releases/tag/v2.0.0), which use Northstar's `v2/` APIs.

🙈 I've removed some usage of "sensitive" fields, like `mobile` and `last_name` that didn't seem necessary in 045e39b. These are no longer returned in the standard response in Northstar's `v2/` APIs & usage should be avoided if possible.

This is part of our "Q3 Upgrades" epic, and ensures that we keep getting security and bug fixes! It also includes new features, so it's worth skimming the [Laravel 5.8 Release Notes](https://laravel.com/docs/5.8/releases#laravel-5.8) and [Laravel 6.0 Release Notes](https://laravel.com/docs/6.x/releases#laravel-6) to see what's new! 🆕 

### Relevant tickets

References [Pivotal #172383688](https://www.pivotaltracker.com/story/show/172383688).

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/ImportFileController.php
+++ b/app/Http/Controllers/Web/ImportFileController.php
@@ -47,14 +47,19 @@ class ImportFileController extends Controller
 
         if ($importType === ImportType::$rockTheVote) {
             $userId = \Auth::user()->northstar_id;
-            $user = gateway('northstar')->asClient()->getUser($userId);
+            $user = gateway('northstar')->asClient()->getUser($userId, [
+                'addr_street1', 'addr_street2', 'email', 'mobile', 'last_name',
+            ]);
 
             $data = [
-                'display_name' => $user->display_name,
+                'addr_street1' => $user->addr_street1,
+                'addr_street2' => $user->addr_street2,
                 'addr_city' => $user->addr_city,
                 'addr_zip' => $user->addr_zip,
-                'email_preview' => $user->email_preview,
-                'phone_preview' => $user->mobile_preview,
+                'email' => $user->email,
+                'first_name' => $user->first_name,
+                'last_name' => $user->last_name,
+                'phone' => $user->mobile,
                 'tracking_source' => 'source:test,source_details:ChompyUI',
                 'started_registration' => Carbon::now()->format('Y-m-d H:i:s O'),
             ];
--- a/app/Http/Controllers/Web/ImportFileController.php
+++ b/app/Http/Controllers/Web/ImportFileController.php
@@ -44,19 +44,17 @@ class ImportFileController extends Controller
         }
 
         $data = [];
-        $userId = \Auth::user()->northstar_id;
-        $user = gateway('northstar')->asClient()->getUser($userId);
 
         if ($importType === ImportType::$rockTheVote) {
+            $userId = \Auth::user()->northstar_id;
+            $user = gateway('northstar')->asClient()->getUser($userId);
+
             $data = [
-                'addr_street1' => $user->addr_street1,
-                'addr_street2' => $user->addr_street2,
+                'display_name' => $user->display_name,
                 'addr_city' => $user->addr_city,
                 'addr_zip' => $user->addr_zip,
-                'email' => $user->email,
-                'first_name' => $user->first_name,
-                'last_name' => $user->last_name,
-                'phone' => $user->mobile,
+                'email_preview' => $user->email_preview,
+                'phone_preview' => $user->mobile_preview,
                 'tracking_source' => 'source:test,source_details:ChompyUI',
                 'started_registration' => Carbon::now()->format('Y-m-d H:i:s O'),
             ];
--- a/app/Http/Controllers/Web/ImportFileController.php
+++ b/app/Http/Controllers/Web/ImportFileController.php
@@ -45,7 +45,7 @@ class ImportFileController extends Controller
 
         $data = [];
         $userId = \Auth::user()->northstar_id;
-        $user = gateway('northstar')->asClient()->getUser('id', $userId);
+        $user = gateway('northstar')->asClient()->getUser($userId);
 
         if ($importType === ImportType::$rockTheVote) {
             $data = [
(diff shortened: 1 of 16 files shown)


DFurnes:
@aaronschachter Pinging you for an extra set of eyes since you're probably the most familiar with these jobs!

In particular, I'm curious if you can think of any places where Northstar's new `v2/` APIs (which don't include [sensitive fields](https://github.com/DoSomething/northstar/blob/e83cabc46ae058519106894616ce1a67e777cef1/app/Models/User.php#L167-L181) like `email`, `mobile`, `addr_street1`, etc) might cause us trouble here. It seems like we only read these values on the incoming CSV, so _should_ be fine but let me know if anything comes to mind!

app/Http/Controllers/Web/ImportFileController.php
```diff
@@ -44,19 +44,17 @@ public function create($importType)
...
-                'addr_street2' => $user->addr_street2,
+                'display_name' => $user->display_name,
                 'addr_city' => $user->addr_city,
                 'addr_zip' => $user->addr_zip,
-                'email' => $user->email,
-                'first_name' => $user->first_name,
-                'last_name' => $user->last_name,
-                'phone' => $user->mobile,
+                'email_preview' => $user->email_preview,
+                'phone_preview' => $user->mobile_preview,
```
aaronschachter, line 57: This makes the Test Import form slightly less convenient, since our logged in user would need to manually enter all of their user information. In general though, this is mostly used by me to test adding posts for new users, so I'll append a datetime to my email, and clear out the mobile (otherwise Chompy will find my user by mobile, and creating new user won't work)

The biggest inconvenience however would be the changes that remove the PII from the import result in `ImportRockTheVoteRecord.php`. We use these to display the user info after submitting the test form:
https://github.com/DoSomething/chompy/blob/9a7e8c13b154ee0a413e5f542a05e6d2dd78369c/app/Http/Controllers/Web/ImportFileController.php#L179-L186

You'd have to remember which unique email you used in order to update a post.

This PII is only used for the internal use of this test import form.. what do you think about keeping it in, per the use cases above?

DFurnes: Ah, that's super helpful to know! Yeah, I'm good with keeping this in for that use-case.

DFurnes: I've re-added these fields & updated the Northstar API call to include them in 76c08f3.

app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
```diff
@@ -126,7 +126,7 @@ private function formatResponse(NorthstarUser $user, array $post)
             ]),
         ];
 
-        $userFields = ['id', 'email', 'mobile', 'voter_registration_status', 'sms_status', 'sms_subscription_topics', 'email_subscription_status', 'email_subscription_topics', 'referrer_user_id'];
+        $userFields = ['id', 'email_preview', 'mobile_preview', 'voter_registration_status', 'sms_status', 'sms_subscription_topics', 'email_subscription_status', 'email_subscription_topics', 'referrer_user_id'];
```
aaronschachter, line 129: I think we need to make these changes too per https://github.com/DoSomething/chompy/pull/177/commits/76c08f3012b5e97ea52ff76c45606c563fafac81

DFurnes: Oh, yep! Done in 1a7a800.

DFurnes: Although, I'm realizing now that those fields will not be included in the [`getUserByEmail`](https://github.com/DoSomething/chompy/pull/177/files#diff-cd7de09fd28f032192b6f11e74524a0dR165) or [`getUserByMobile`](https://github.com/DoSomething/chompy/pull/177/files#diff-cd7de09fd28f032192b6f11e74524a0dR178) calls unless we specifically ask for them (and so we'd have to either request them across the board, or perhaps pass another `$debug` parameter to the job to tell it whether or not to query those...)

DFurnes: I'm wondering if printing the provided form values (from the `$record` passed to this job), and non-sensitive results (just a boolean "found email user" or "found mobile user" or the IDs for each) would provide us enough context for debugging?

aaronschachter: I think printing the form values should work! That form submission flash message currently doesn't indicate whether user/post was updated/created (would be sweet if it did -- but I'd imagine that's just a nice-to-have since we don't do that)

aaronschachter: I don't think it's a biggie to lose pre-populating the form with current user info (it's convenient to have the email ready to be edited, but also very easy to clear out your mobile - which prevents creating a new user)

aaronschachter:
Testing this out locally, and running into `Class 'Illuminate\Support\Facades\Input' not found` over in 
https://github.com/DoSomething/chompy/blob/586b32b905498d89abc2cde7e3d124bf0b3d7d52/app/Http/Controllers/Web/ImportFileController.php#L188

when using the Test Import form.

DFurnes:
Strange! I can't imagine how that was working in the first place. 🤔 I can fix up in my follow-up PR.

PR #179, opened 2020-07-15
Print record alongside user, mobile user, and post.

DFurnes:
### What's this PR do?

This pull request updates the "debug results" that we return from the `ImportRockTheVoteRecord@handle` method to be a bit more verbose, now that we don't include the `email` and `mobile` fields in the base response.

Here's an example from my testing (where it resolved my main user, and a mobile-only test user):

Screen Shot 2020-07-15 at 1 44 07 PM

👀 

This addresses https://github.com/DoSomething/chompy/pull/177#discussion_r453897512. Because we don't return PII in the `v2/users` response objects, we can no longer easily log the `email` and `mobile` fields here. (And because these `$user` and `$this->mobileUser` values are queried for _all_ jobs, we don't want to always request those fields since it'd add a ton of noise to our logs.)

To make it as easy as possible to debug jobs locally, this form now prints out the `$this->record` input values (from the form, which includes any provided PII) alongside the "safe" `$user` and `$this->mobileUser` and `$post`.

### Relevant tickets

References [Pivotal #172383688](https://www.pivotaltracker.com/story/show/172383688).

### Checklist

### DIFF
--- a/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
+++ b/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
@@ -23,9 +23,16 @@ class ImportRockTheVoteRecord implements ShouldQueue
     /**
      * The record parsed from a Rock the Vote csv.
      *
-     * @var array
+     * @var RockTheVoteRecord
      */
     protected $record;
+
+    /**
+     * The mobile user (if different from "primary" user).
+     * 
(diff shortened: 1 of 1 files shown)


DFurnes:
Note: I fixed the StyleCI issues, but GitHub's taking it's time with the new commit! 🙈

PR #182, opened 2020-07-20
Retry importing a failed Rock The Vote report

aaronschachter:
### What's this PR do?

This pull request adds a "Retry" button to the UI when viewing a RTV report that has a `failed` status, and hasn't been retried yet. Under the hood, this happens per the following changes:

* Adds DB migration / model attribute for a new `retry_report_id` int column on the `rock_the_vote_reports` table

* Adds a `createRetryReport` method to the `RockTheVoteReport` model. This method uses the report's `since` and `before` parameters to execute an RTV API request to create a new report, and sets the new ID from API response to its `retry_report_id`.

* Adds the `update` route and handler for the `RockTheVoteReport` resource, which will call `createRetryReport` for a given `RockTheVoteReport` id

Example retry form:

Example retried report:

👀 

* This PR also adds a `faker` configuration key for the Rock The Vote service, which I wanted in order to make and test the UI updates without sending real API requests to the RTV, which would unnecessarily generate reports. 

* I plan to add additional test coverage for the update form in a future PR (verify when Retry button is visible, that import job is dispatched). Left it out for now in hopes of keeping this easier to review

### Relevant tickets

### Checklist

### DIFF
--- a/app/Models/RockTheVoteReport.php
+++ b/app/Models/RockTheVoteReport.php
@@ -89,4 +89,19 @@ class RockTheVoteReport extends Model
     {
         return round(($this->current_index * 100) / $this->row_count);
     }
+
+    /**
+     * Creates a new report with this report's time and since, and saves it to retry_report_id.
+     *
+     * @return RockTheVoteReport
+     */
+    public function createRetryReport()
+    {
+        $retryReport = self::createFromApi($this->since, $this->before);
+
+        $this->retry_report_id = $retryReport->id;
+        $this->save();
+
+        return $retryReport;
+    }
 }
--- a/app/Models/RockTheVoteReport.php
+++ b/app/Models/RockTheVoteReport.php
@@ -18,13 +18,14 @@ class RockTheVoteReport extends Model
      * @var array
      */
     protected $fillable = [
-        'id',
-        'status',
-        'since',
         'before',
-        'row_count',
         'current_index',
         'dispatched_at',
+        'id',
+        'retry_report_id',
+        'row_count',
+        'since',
+        'status',
         'user_id',
     ];
 
@@ -34,9 +35,9 @@ class RockTheVoteReport extends Model
      * @var array
      */
     protected $dates = [
-        'since',
         'before',
         'dispatched_at',
+        'since',
     ];
 
     /**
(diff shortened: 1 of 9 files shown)


app/Models/RockTheVoteReport.php
```diff
@@ -87,4 +105,19 @@ public function getPercentageAttribute()
...
+    /**
+     * Creates a new report from since and before, and saves it to retry_report_id.
+     *
+     * @return RockTheVoteReport
+     */
+    public function createRetryReport()
+    {
+        $retryReport = self::createViaApi($this->since, $this->before);
+
+        $this->retry_report_id = $retryReport->id;
```
mendelB, line 118: If the retry also fails, is the idea that _it_ in turn would display it's own retry button which would lead to _it's_ retry ID -> thus creating a navigable chain?

tests/Unit/Models/RockTheVoteReportTest.php
```diff
@@ -8,6 +8,33 @@
...
+    protected $isFakerEnabled;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        $this->isFakerEnabled = config('services.rock_the_vote.faker');
+
+        if ($this->isFakerEnabled) {
+            \Config::set('services.rock_the_vote.faker', false);
```
mendelB, line 25: Do I understand this correctly that we turn this off for testing because we want to run through the actual functionality - which uses the mocks anyhow, bypassing the actual API?

app/Models/RockTheVoteReport.php
aaronschachter: Yep! Not the most convenient way to navigate a failed job's journey, but we're looking for minimal footprint on this feature.

tests/Unit/Models/RockTheVoteReportTest.php
aaronschachter: We want to unset the "faker" flag that's used for manually testing the UI without sending actual API requests to RTV. We don't have access to a dev or QA instance of the Rock the Vote API. I wanted to add this to at least see with my own 👀 that the form worked correctly beyond just relying on the automated tests (it's burned me before).

PR #183, opened 2020-07-22
Retry failed reports via ImportRockTheVoteReport job

aaronschachter:
### What's this PR do?

This pull request continues building out the retry feature introduced in #182, modifying the `ImportRockTheVoteReport` job to better handle when its report status is `failed`:

* If the report hasn't been retried, create a retry report and dispatch the new retry report ID for import, discarding the current job for the failed report.

* If the report has been retried, discard the current job for the failed report. The only time we'd possibly run into this is when an admin manually creates a retry via UI before the hourly import runs to create the retry.

:eyes:

🦎 

### Relevant tickets

### Checklist

### DIFF
--- a/tests/Jobs/RockTheVote/ImportRockTheVoteReportTest.php
+++ b/tests/Jobs/RockTheVote/ImportRockTheVoteReportTest.php
@@ -13,11 +13,11 @@ use Chompy\Jobs\ImportRockTheVoteReport;
 class ImportRockTheVoteReportTest extends TestCase
 {
     /**
-     * Test that report is not downloaded when status not complete.
+     * Test that report is not downloaded and job is retried if status is building.
      *
      * @return void
      */
-    public function testReportStatusNotComplete()
+    public function testReportStatusBuilding()
     {
         Bus::fake();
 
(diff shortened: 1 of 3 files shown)


tests/Jobs/RockTheVote/ImportRockTheVoteReportTest.php
```diff
@@ -13,11 +13,11 @@
 class ImportRockTheVoteReportTest extends TestCase
 {
     /**
-     * Test that report is not downloaded when status not complete.
+     * Test that report is not downloaded and job is retried if status is building.
```
katiecrane, line 16: I thought it was only supposed to be retried if it failed?

aaronschachter: Ahh this is a confusing comment on my end. We dispatch a job to check this report status again after 2 minutes since it's still building. Maybe this should just read as 
```
Test that a job is dispatched to import this report again after 2 minutes if its status is building
```

mendelB: I like this alternative!

PR #184, opened 2020-07-24
Feature flag automatic retry of failed reports

aaronschachter:
### What's this PR do?

This pull request checks to see if a new config variable, `ROCK_THE_VOTE_RETRY_FAILED_REPORTS_ENABLED` is set to `true` before retrying a failed report via console command (introduced in #183).

We've seen that retrying certain reports' `since` and `before` parameters returns a 500 error from the RTV API, so we rolled back the deploy of #183. TBD whether we'll end up setting `ROCK_THE_VOTE_RETRY_FAILED_REPORTS_ENABLED` at all - but figured the feature flag wouldn't hurt for now vs. deleting all the code and tests that #183 introduced.

👀 

I tried restoring the default configuration in `tearDown`, like we're doing for #182 -- but the tests fail with:
```
Illuminate\Contracts\Container\BindingResolutionException: Target class [config] does not exist.
```
and
```
ReflectionException: Class config does not exist
```

I still can't figure out why this is happening, but got it working by setting the `config` back in each individual test for now 😤 

### Relevant tickets

### Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -86,6 +86,9 @@ NEWS_SUBSCRIPTION_RESET_TYPE=breakdown-activate-account
 ## Optional config overrides - Rock The Vote Import
 ##
 
+# Whether to retry a failed report when importing from console command. 
+ROCK_THE_VOTE_RETRY_FAILED_REPORTS_ENABLED=false
+
 # Whether to update existing users' SMS subscriptions per opt-in via RTV registration form.
 ROCK_THE_VOTE_UPDATE_USER_SMS_ENABLED=false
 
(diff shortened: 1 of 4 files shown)


DFurnes:
> I tried restoring the default configuration in `tearDown`, like we're doing for #182 -- but the tests fail with:
> 
> ```
> Illuminate\Contracts\Container\BindingResolutionException: Target class [config] does not exist.
> ```
> 
> and
> 
> ```
> ReflectionException: Class config does not exist
> ```
> 
> I still can't figure out why this is happening, but got it working by setting the `config` back in each individual test for now triumph

Hmm, that's so strange! If you still have the code stashed away, could you maybe push it up after merging this? The "class does not exist" sounds like an application container issue, so maybe something with it being unable to instantiate the `Config` manager after the app has been torn down?

If we've got things set up correctly (we might not), you shouldn't have to "reset" things between tests since each test should be run with a brand new freshly instantiated application container, so that's also worth digging into a bit if that's not working.

PR #191, opened 2020-10-05
Do not add phone number to users who did not opt in to receive SMS messages

katiecrane:
### What's this PR do?

This pull request updates some fun business logic. We only want to save a user's phone number to a new or existing account IF the user has opted in to receiving SMS messaging. Previously in these cases, we would save the phone number and set their `sms_status` to `stop` but that was confusing when looking at the data because it looks like these folks have actively unsubscribed, when really they were never subscribed in the first place.

The only change this should make should be to not save mobile number to a user's profile if they have not opted in to sms. Did I update tests properly? Should I write any additional tests?

Check out the thread on the pivotal ticket for more context!

### Relevant tickets

References [Pivotal #174361978](https://www.pivotaltracker.com/story/show/174361978).

### Checklist

### DIFF
--- a/app/RockTheVoteRecord.php
+++ b/app/RockTheVoteRecord.php
@@ -58,15 +58,19 @@ class RockTheVoteRecord
                 'email_subscription_topics' => $emailOptIn ? explode(',', $config['user']['email_subscription_topics']) : [],
                 'first_name' => $record['First name'],
                 'last_name' => $record['Last name'],
-                'mobile' => isset($record[static::$mobileFieldName]) && is_valid_mobile($record[static::$mobileFieldName]) ? $record[static::$mobileFieldName] : null,
             ]);
 
-            // If a mobile was provided, set SMS subscription per opt-in value.
-            if ($this->userData['mobile']) {
-                $smsOptIn = str_to_boolean($record[static::$smsOptInFieldName]);
+            // If a the user opted in to SMS messaging, add their mobile number
+            $smsOptIn = str_to_boolean($record[static::$smsOptInFieldName]);
+            if ($smsOptIn) {
+                $this->userData['mobile'] = isset($record[static::$mobileFieldName]) && is_valid_mobile($record[static::$mobileFieldName]) ? $record[static::$mobileFieldName] : null;
 
-                $this->userData['sms_status'] = $smsOptIn ? SmsStatus::$active : SmsStatus::$stop;
-                $this->userData['sms_subscription_topics'] = $smsOptIn ? explode(',', $config['user']['sms_subscription_topics']) : [];
+                // Only set SMS subscription fields if mobile is included and valid
+                if($this->userData['mobile']) {
+                    $this->userData['sms_status'] = $smsOptIn ? SmsStatus::$active : SmsStatus::$stop;
+                    $this->userData['sms_subscription_topics'] = $smsOptIn ? explode(',', $config['user']['sms_subscription_topics']) : [];
+                }
+
             }
         }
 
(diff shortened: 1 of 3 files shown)


DFurnes (approved):
Makes sense to me! 📴

app/RockTheVoteRecord.php
```diff
@@ -58,15 +58,18 @@ public function __construct($record, $config = null)
...
-            // If a mobile was provided, set SMS subscription per opt-in value.
-            if ($this->userData['mobile']) {
-                $smsOptIn = str_to_boolean($record[static::$smsOptInFieldName]);
-
-                $this->userData['sms_status'] = $smsOptIn ? SmsStatus::$active : SmsStatus::$stop;
-                $this->userData['sms_subscription_topics'] = $smsOptIn ? explode(',', $config['user']['sms_subscription_topics']) : [];
+            // If a the user opted in to SMS messaging, add their mobile number and SMS subscription fields
+            $smsOptIn = str_to_boolean($record[static::$smsOptInFieldName]);
+            if ($smsOptIn) {
+                if (isset($record[static::$mobileFieldName]) && is_valid_mobile($record[static::$mobileFieldName])) {
```
aaronschachter, line 68: This is nitpicky and optional, but I've been trying to keep line breaks and conditionals per our [communal docs](https://github.com/DoSomething/communal-docs/tree/master/Contributing#line-breaks) when adding/updating code in this repo. I think a single conditional could help simplify?

```suggestion
            $smsOptIn = str_to_boolean($record[static::$smsOptInFieldName]);

            if ($smsOptIn && isset($record[static::$mobileFieldName]) && is_valid_mobile($record[static::$mobileFieldName])) {
```

aaronschachter (approved):
I think this looks good. Just sanity checking, have you manually tested via /import/rock-the-vote?source=test to spot check? I've been burned before on just relying on unit tests and mocks.

aaronschachter (changes requested):
I forgot about an important part of the repo we'll want to update per this change, [documentation](https://github.com/DoSomething/chompy/tree/master/docs/imports#mobile).

This feature does mean that if a subscribed user happens to have the `voting` SMS topic, opting out will no longer remove the topic -- which was part of the initial requirements when we built out updating an existing user's SMS preference.

aaronschachter:
The docs on [SMS subscription](https://github.com/DoSomething/chompy/tree/master/docs/imports#sms-subscription) also need updates.

app/RockTheVoteRecord.php
katiecrane: ah good call on the spacing! I like combining them too, previously I had this logic less streamlined but now could totally do all in one line like that!

katiecrane:
@aaronschachter I did my initial testing while I was working on it with /import/rock-the-vote?source=test and then moved on to fixing failing tests/adding new ones from there! GREAT CALL on the documentation updates, I totally spaced on that! 

Ah I did not know about that voting topic requirement - is this what should be happening?
1. we have an existing user who has `sms_status` of `active` and `voting` as one of their `sms_subscription_topics`
2. they come through the RTV flow, provide their phone number, but do not opt-in to SMS messages
3. they keep the same `mobile`, `sms_status` that they already have on their profile, but `voting` is removed from their `sms_subscription_topics`

vs. what is happening with my changes:
1. we have an existing user who has `sms_status` of `active` and `voting` as one of their `sms_subscription_topics`
2. they come through the RTV flow, provide their phone number, but do not opt-in to SMS messages
3. we make no changes to their `mobile`, `sms_status`, or `sms_subscription_topics`

Let me know if it'd be helpful to hop on a call to discuss!

weerd:
@katiecrane I'm going to close this PR since we're going to archive this repository since the codebase has been merged with Northstar! Is this something we should address in Northstar at a later date? If so I can make a ticket and link to this closed PR for reference and code approach 😄

PR #192, opened 2020-10-22
Validate RTV "Started registration" value

aaronschachter:
### What's this PR do?

This pull request adds a check to only import a RTV record if its `Started registration` value is a date.

👀 

A great follow-up cleanup task would be to use this as the groundwork for better validation of the `Phone number` field before we import it. We use a very basic [helper](https://github.com/DoSomething/chompy/blob/4e253ddcb2b1f6056468c5e64b50490d57b253f8/app/helpers.php#L50)... but it still doesn't catch [all variations](https://dosomething.slack.com/archives/CTVPG6L4R/p1603215583087800?thread_ts=1603215534.087600&cid=CTVPG6L4R)

### Relevant tickets

References [Pivotal #175256802](https://www.pivotaltracker.com/story/show/175256802).

### Checklist

### DIFF
--- a/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
+++ b/app/Jobs/RockTheVote/ImportRockTheVoteRecord.php
@@ -14,6 +14,7 @@ use Chompy\Models\RockTheVoteLog;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Validation\ValidationException;
 use DoSomething\Gateway\Resources\NorthstarUser;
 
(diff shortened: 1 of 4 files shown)


DFurnes (approved):
Nice! 🍡

DFurnes:
> A great follow-up cleanup task would be to use this as the groundwork for better validation of the `Phone number` field before we import it. We use a very basic [helper](https://github.com/DoSomething/chompy/blob/4e253ddcb2b1f6056468c5e64b50490d57b253f8/app/helpers.php#L50)... but it still doesn't catch [all variations](https://dosomething.slack.com/archives/CTVPG6L4R/p1603215583087800?thread_ts=1603215534.087600&cid=CTVPG6L4R)

Yeah, that would be great. I wonder what other fields Rock The Vote might give us weird/inconsistent data for too... 🤔

PR #202, opened 2021-02-24
Use new Northstar endpoint to mute promotions

aaronschachter:
### What's this PR do?

This pull request alters the `ImportMutePromotions` job to execute a Northstar API request to the new `POST /users/:userId/promotions` endpoint added in https://github.com/DoSomething/northstar/pull/1132, instead of `PUT /users/:userId`.

This allows us to successfully set the `promotions_muted_at` field for users who may have existing null `email` and `mobile` field values (e.g. GDPR users).

👀 

🌵 

### Relevant tickets

References [Pivotal #176771195](https://www.pivotaltracker.com/story/show/176771195).

### Checklist

- [x] Documentation added for new features/changed endpoints.

### DIFF
--- a/app/Jobs/MutePromotions/ImportMutePromotions.php
+++ b/app/Jobs/MutePromotions/ImportMutePromotions.php
@@ -2,6 +2,7 @@
 
 namespace Chompy\Jobs;
 
+use Exception;
 use Chompy\Models\ImportFile;
 use Chompy\Models\MutePromotionsLog;
 use Illuminate\Bus\Queueable;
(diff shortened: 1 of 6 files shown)


app/Jobs/MutePromotions/ImportMutePromotions.php
```diff
@@ -35,22 +34,18 @@ public function __construct($record, ImportFile $importFile)
...
-     *
-     * @return array
+     * Execute the job to mute user promotions.
      */
     public function handle()
     {
-        $user = gateway('northstar')->asClient()->updateUser($this->userId, [
-            'promotions_muted_at' => Carbon::now(),
-        ]);
+        gateway('northstar')->asClient()->delete('v2/users/'. $this->userId . '/promotions');
```
aaronschachter, line 41: The return value here is a boolean, so I changed this up.

DFurnes (approved):

app/Jobs/MutePromotions/ImportMutePromotions.php
DFurnes: In Gateway, we return `true` or `false` for delete requests to indicate whether they were successful or not. In hindsight, I'm not sure this was the best API decision to make across the board like this but it's what we've got! 😅

aaronschachter: Ah, I see. I think we may want to change this slightly to throw an error in this function if the response is `false`, so we can retry any failed jobs (or inspect why they failed)

aaronschachter: I see that Gateway is looking to see a `success` property in [the API response](https://github.com/DoSomething/gateway/blob/dec50f36ecdbd1023e7346c0e261dcb4e81ca5dd/src/Common/RestApiClient.php#L313-L316) of a delete request when determining the return value. It feels kinda icky but I suppose we can modify the `DELETE /users/:id/promotions` response to return both `data` and `success` properties so we'll be able to know when requests fail?

e.g.
```
{
    "success": true,
    "data": {
        "id": "600201a023a8223a1e4575a7",
        "display_name": "Virgie D.",
        "first_name": "Virgie",
        "last_initial": "D",
        "photo": null,
        ...
}
````

File: resources/views/pages/partials/mute-promotions/logs.blade.php [hunk 2]
```diff
@@ -15,3 +15,5 @@
         </tr>
     @endforeach
 </table>
+
+{{$rows->links()}}
```
aaronschachter, line 19: This throws an error `Method Illuminate\Database\Eloquent\Collection::links does not exist.` when being rendered from the user view.

PR #201, opened 2021-02-22
Adds log table for mute promotions import

aaronschachter:
### What's this PR do?

This pull request continues work done in #200 to add a `MutePromotionsLog` model, which may help troubleshoot any imports we run once we go live (storing each imported user in a DB table vs needing to query Papertrail logs).

It also stores the original uploaded file name into the `options` serialized field, to make it easier to keep track of which of the [10 CSV's](https://trello.com/c/8P4yy1Ys/1734-csvs-w-profiles-to-remove-from-customerio) have been uploaded for import so far.

### How should this be reviewed?

Import File view:

User view:

### Any background context you want to provide?

We discussed in the [Pivotal thread](https://www.pivotaltracker.com/n/projects/2328687/stories/176771195/comments/222168919) that we're not too worried about storage implications this may have for now (we'll be running this import for millions of users this week).

### Relevant tickets

References [Pivotal #176771195](https://www.pivotaltracker.com/story/show/176771195).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [x] Documentation added for new features/changed endpoints.
- [ ] Added appropriate feature/unit tests.

PR #205, opened 2021-02-26
Fixes error on user view

aaronschachter:
### What's this PR do?

This pull request adds a check to see whether the `links` method exists on the `$rows` variable within the Mute Promotions logs added in #201. We shouldn't be uploading too many CSV's that repeat user ID's, which is why this view shouldn't need to paginate when viewing a user. The pagination links are more likely to be used on the Rock The Vote section, which appears below the Mute Promotions logs.

This could probably be cleaned up to have different routes per import, e.g. `user/:id/:importType` but for now this will avoid the server error rendered when viewing a user (e.g. `/user/:id`)

👀 

🚑 

### Relevant tickets

References [Pivotal #176771195](https://www.pivotaltracker.com/story/show/176771195).

### Checklist

### DIFF
--- a/resources/views/pages/partials/mute-promotions/logs.blade.php
+++ b/resources/views/pages/partials/mute-promotions/logs.blade.php
@@ -16,4 +16,6 @@
     @endforeach
 </table>
 
-{{$rows->links()}}
+@if(method_exists($rows, 'links'))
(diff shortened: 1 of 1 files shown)


Dinnall (approved):
LGTM

PR #206, opened 2021-02-26
Add high & low priority queue support.

DFurnes:
### What's this PR do?

This pull request updates Chompy to dispatch file parsing jobs to a separate high priority queue, which workers will read from before processing anything else. This follows a similar approach to how we do this in Northstar, except I've chosen to default to the low-priority queue (since that's the usual case, in this app).

I based this on Northstar's [Procfile](https://github.com/DoSomething/northstar/blob/master/Procfile), [config/queue.php](https://github.com/DoSomething/northstar/blob/master/config/queue.php), and [example usage](https://github.com/DoSomething/northstar/blob/ac7b1147bee522cad10c9914ff3ab01a63c27f6e/app/Console/Commands/BackfillCustomerIo.php#L38-L40).

Before merging this, I'll create these new "high-priority" SQS queues, attach IAM permissions, and set the required `SQS_HIGH_PRIORITY_QUEUE` environment variables on QA & production.

We're running into issues where the `ImportFileRecords` job is being processed multiple times. We [think this is because SQS's visibility timeout](https://dosomething.slack.com/archives/C02BBP0CU/p1614293344007500?thread_ts=1614283088.000700&cid=C02BBP0CU) is allowing multiple workers to try their hand at this long-running job (and therefore processing each row 3 or more times, blah)!

As a workaround, we'd like to upload all of our files while using a _single_ queue worker & only scale up to parallel workers once we've got all the downstream processing jobs in the queue. However, by default records are processed single file (e.g. import file, process all of it's records, then import the next file...).

We'll add a new "high priority" queue for `ImportFileRecords` so those always take priority over other jobs that are in the standard queue, allowing us to run all these imports before we scale up and process their rows!

### Relevant tickets

References [Pivotal #177118983](https://www.pivotaltracker.com/story/show/177118983).

### Checklist


DFurnes:
We were also considering using SQS's [FIFO queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/welcome.html#sqs-queue-types) to ensure that we couldn't process the same job more than once. Unfortunately, Laravel only seems to support "normal" SQS queues out of the box, and we'd need a [third-party package](https://github.com/shiftonelabs/laravel-sqs-fifo-queue) to support FIFO queues. I'd like to hold on that for now so we're not making too many changes at once.

DFurnes:
Oh interesting – we've also set a visibility timeout of 0 seconds right now on the `chompy` and `chompy-qa` queues...

> Visibility timeout sets the length of time that a message received from a queue (by one consumer) will not be visible to the other message consumers.
>
> The default visibility timeout setting is 30 seconds. This setting applies to all messages in the queue. Typically, you should set the visibility timeout to the maximum time that it takes your application to process and delete a message from the queue.

I'm going to set this to `720` for the new high priority queue & `30` for the low-priority queue.

DFurnes:
Alright, these queues should be set up (this app isn't in Terraform yet, so configured by hand). Let's test out a couple file uploads in QA first to see if everything seems to work as expected & then we can try using this in production!

DFurnes:
I tried uploading the first three CSVs on QA & confirmed that three jobs end up in the `chompy-qa-high` queue. The queue worker is currently processing the first file upload & placing jobs into the `chompy-qa` queue. 👍

The file import itself seems to be running quite slow (only 20,876 rows after 10 minutes), but I wouldn't think that's related...

aaronschachter (approved):

DFurnes:
Cleared out the QA queue since that was taking so long & trying again with three tiny 1000-line CSVs.

DFurnes:
That was a way better test! Confirmed that all three import file jobs were processed first before processing rows.

aaronschachter:
> That was a way better test! Confirmed that all three import file jobs were processed first before processing rows.

Awesome! So it sounds like the order of operations for today will be:

* Deploy this, set Chompy worker dynos to 1
* Upload the other 19 CSV's
* When they are complete, bump up Chompy dynos to 6+

DFurnes:
I think so! I'm curious though whether we might be able to process file imports in parallel now that we've got the right visibility timeout setting... I might try testing that with these three tiny CSVs on QA to see how it goes.

DFurnes:
Good news - we got exactly the right number of records imported for each of the three CSVs on QA (the first one is 998 because I split the file with `split -n 1000` and it doesn't know to take the header row into account):

Screen Shot 2021-02-26 at 12 50 57 PM

I'm going to temporarily scale Chompy QA up to 6 queue workers & see how they handle the same task in parallel.

DFurnes:
Mostly good news! Somehow one of the files got parsed twice, but otherwise everything seems to have been processed exactly as expected (and much, much faster). I think this is worth trying in production & seeing how it goes!

Screen Shot 2021-02-26 at 12 57 17 PM

DFurnes:
Scaled Chompy QA down to a single queue worker.

